PR #17, opened 2020-04-29
Python2 and Python3 compatible

volkerjaenisch:
Dear John!

It took a while but now the code runs flawlessly under python2 and python3 on the same code base, without if constructs.
One crucial part was to restructure the test data in the JSON files. Namely splitting the payload from the CRC.
The background is 
[https://tools.ietf.org/id/draft-ietf-json-rfc4627bis-09.html#rfc.section.8.2](https://tools.ietf.org/id/draft-ietf-json-rfc4627bis-09.html#rfc.section.8.2)
which states:
`However, the ABNF in this specification allows member names and string values to contain bit sequences which cannot encode Unicode characters, for example "\uDEAD" (a single unpaired UTF-16 surrogate). Instances of this have been observed, for example when a library truncates a UTF-16 string without checking whether the truncation split a surrogate pair. The behavior of software which receives JSON texts containing such values is unpredictable; for example, implementations might return different values for the length of a string value, or even suffer fatal runtime exceptions.
`
So there seems no reliable way to store binary coded information in json beside hex strings. So I decided to separate the CRC as a hex string from the payload.

Cheers,
Volker


ceaza:
@volkerjaenisch  thanks for making this project Python3 compatible.

PR #29, opened 2020-06-22
Esp32 support

volkerjaenisch:
The ESP32 Code is stil under development. Nevertheless I like to have the branch included in the base repo to prevent more divergence due to your really active development.


jblance:
Looking pretty good.
I think the combine the json into a single .py file is a good idea, but would prefer to do it in a way that supports the different protocols better - it might be quicker for you to make the changes?
Otherwise I can merge this into main and change it.

Essentially I am thinking of a file per protocol (have started looking at this in a different repo). So the standard pi30.py protocol would have all the commands supported by the 'standard' and we have a different file (with the same interface) for different protocols.

So would run the convert tool of yours and remove the json files (and tool) completely, but have a file for each different protocol.

check out the powerman repo (which is very incomplete) to see what I am thinking.

volkerjaenisch:
I would really like if you do the per protocol transformation. You have for sure good goals and I may do more harm than good.

In our automotive project we have some kind of layered configuration for different protocols/configs. In most cases we have a base protocol/config and then incremental changes which are mostly minor. At first this was an overkill but as the number of machine series increased it made our lives and that of their engineers much easier. Sounds like you have something similar in mind.

My role is and will be the esp32 port and the testing of it. At the weekend I will try to build a serial to serial bridge over a VPN to my customer. The latency is 50ms which may be tolerable for such a slow serial connection.
Goal is to be able to test comfortably here in my office and do not have to flash the ESP remotely which drags the turnaround time at infimum.

jblance:
I've got the start of this 'per protocol' in the refactor branch
It is supposed to be a JIT import of the relevant 'modules' for device, IO communications, protocol and outputs
Can you see if this would work for you?

jblance:
I think I got most of this into the refactor that I have done
I will be testing on esp32 soon
Closing this as code as diverged significantly since

PR #37, opened 2020-08-11
extrapolates real path from device symlink + small bug fix

mholgatem:
I am recommending making some changes to reading the device path so that udev created symlinks can be used. With these changes, a udev rule can be made in this fashion
```
ACTION=="add", ATTRS{idVendor}=="<vendor>", ATTRS{idProduct}=="<product>", SYMLINK+="mppsolar/direct/0", MODE="0660", GROUP="plugdev"
```
This allows for a more consistent naming scheme for devices, without actually changing their default nodes.

I have created an installer that, among other things, will create a custom udev rule for selected devices. This will create symlinks in the  form of /dev/mppsolar/<connection_type/<id>. connection_type will be either *direct* or *serial*

If you care to check out the installer, it is at https://github.com/mholgatem/mpp-solar-installer
currently it is cloning my fork of your mpp-solar project because of udev changes

### DIFF
--- a/mppsolar/__init__.py
+++ b/mppsolar/__init__.py
@@ -1,5 +1,6 @@
 # !/usr/bin/python3
 import logging
+import os
 from argparse import ArgumentParser
 
 from .version import __version__  # noqa: F401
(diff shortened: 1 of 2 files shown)


jblance:
I specifically left the os import out as some devices failed on it (esp32) - so i have left this out of the code
the rest is incorporated (i think) though in a different way as I have done a large refactor

PR #51, opened 2020-12-28
Fix crcPI Space issue / Removed ctypes import and replaced with &s

Northbadge:
crc overflowed 16 bits, but Python has no overflowing so that broke; Added bitwise ands to limit the bits to uint16

Removed ctypes dependency and replaced with &s

### DIFF
--- a/mppsolar/protocols/protocol_helpers.py
+++ b/mppsolar/protocols/protocol_helpers.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 import logging
-import ctypes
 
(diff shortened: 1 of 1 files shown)


Northbadge:
Quick question, cause idk where to ask, but in `serialio` why use `serial.readline()` instead of `serial.read_until(expected=b'\r')`? The former always times out unless it's `\n`

jblance (approved):

jblance:
No reason other than I wasn't aware of the second approach :-)

jblance:
Thanks for the PR!

PR #52, opened 2021-01-03
Reduce serialio delay, add response crc checking

Northbadge:
Using read_until() for serial reads is faster, since it doesn't timeout. CRC checking responses can be added to individual protocols now as well.

p/s the serialio for loop doesn't appear to... loop? Not sure if that's even needed cause I haven't had trouble with that yet


jblance:
Thanks for the improvements.

Re the non loop, there was a problem previously  but its
1. Not working now
2. Was based on checking for valid response
3. Probably not needed with your change

Will think on it and either remove it completely or reinstate it properly (especially now you've fixed the validity check...)

Cheers
John

PR #58, opened 2021-01-11
Fixed default baud rate setting, add run_commands, separate IO from Protocol (except JKBLE)

Northbadge:
I (mostly) separated IO from protocol since it was (getting) difficult to read through the code. Now IO exclusively sends and returns raw bytes, decoding is done from `devices`. 

I'm not too familiar with JKBLE stuff so I left that mostly untouched, it seems to be rather 'entangled' with the JK protocol file so I didn't know how to deal with it. Perhaps you could look into that? There's a band-aid `isinstance` check for jkble io to use the 'old' method

Also, I wanted to change 
https://github.com/Northbadge/mpp-solar/blob/fa1e552691669a66632ecd3fb5d18b1cdd2cebfe/mppsolar/protocols/protocol.py#L104-L107
to "WARNING" instead of "ERROR"; Your thoughts on that?

### DIFF
--- a/mppsolar/devices/device.py
+++ b/mppsolar/devices/device.py
@@ -135,6 +135,30 @@ class AbstractDevice(metaclass=abc.ABCMeta):
             result[command] = [self._protocol.COMMANDS[command]["description"], ""]
         return result
 
+    def run_commands(self, commands) -> dict:
+        """
+        Run multiple commands sequentially
+        :param commands: List of commands to run, either with or without an alias.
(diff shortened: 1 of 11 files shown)


jblance (approved):
Wow - nice work, thanks for the PR and all your work

jblance:
Re WARNING vs ERROR - I dont really mind, currently the code shouldnt run a command that it doesnt know about (hence ERROR), but it is not a strong preference from me

PR #65, opened 2021-01-31
Fix typo: esp_format -> resp_format

20after4:
Fixes the following error:

Traceback ...
File "...src/mpp-solar/mppsolar/protocols/protocol.py", line 173, in decode
NameError: name 'esp_format' is not defined


jblance:
Thanks for the pr

PR #66, opened 2021-01-31
Handle unexpected / unknown flags

20after4:
The PIP-3048LV-MK returns two unknown QFLAGS which caused the code
in protocol.py to error out with a key error.

This change handles the error condition by checking for the key
in the protocol definition and if it is not found using a fallback
key ("unknown_" followed by the character that represents the flag
in the raw protocol.)


20after4:
I intend to figure out what the unknown flags are and document them, however, this is a stop-gap in the meantime.
This change should make the code more robust and hopefully avoid crashing when other flags are added in future firmware revisions.

20after4:
BTW, thank you for your work on this project. It seems really well designed and implemented.

jblance:
Great, thanks a lot

PR #160, opened 2021-09-30
PI17 update

VictorBucha:
PI17 updated with EY, ED, EM, EH commands
[pi17.zip](https://github.com/VictorBucha/mpp-solar/files/7259030/pi17.zip)


jblance:
Have merged this, but there were a bunch of conflicts (hopefully they were resolved correctl but it would pay to check if i have messed any up....)

PR #193, opened 2022-01-05
Send commands to mpp-solar service via broker

rossandrews:
This is my first pull request to an open source project and my first taste of python so please excuse any major missteps.

My main goals were to add a mechanism to run commands at different rates and send commands to be run by the service. When I tried to execute commands outside of the service they often collided and failed. 

To change the rate commands are run I added the pause_loops option int he config file, it's the number of loops to skip before running.

To listen for commands you can include a commands_topic and any strings send will be included in the commands run. I have a planned improvement so the commands sent also includes a response topic to make it easier for two way communication.

I also refactored the main method a bit to make it a bit more object oriented. Further improvement I think would help would be separating the concepts of device, schedule and command. Device defines the port and protocol. Command defines command and output. Schedule will define which commands are running on which devices and how often.

I have not done extensive testing on features I am not using. I noticed tags didn't work with the json_mqtt output and I may have broken them for other outputs. I didn't test filters.

### DIFF
--- a/docs/service/receive-commands.conf
+++ b/docs/service/receive-commands.conf
@@ -1,0 +1,32 @@
+[SETUP]
+#Pause will now account for the time it takes to run the commands
+pause=5
+mqtt_broker=localhost
+
+#Runs no commands, it waits for requests on the commands_topic
+[Schedule_requests]
+port=/dev/hidraw0
+protocol=PI30
+command=
+commands_topic=green_grid/ross_proto_01/command
+outputs=json_mqtt
+pause_loops=0
+mqtt_output_topic=green_grid/ross_proto_01/requests/
+
+#Gets the stats every other loop (10 seconds)
+[Schedule_stats]
+port=/dev/hidraw0
+protocol=PI30
+command=QPIGS
+outputs=json_mqtt
+pause_loops=1
+mqtt_output_topic=green_grid/ross_proto_01/
+
+#Getings the device settings every 6 loops (30 seconds)
+[Schedule_settings]
+port=/dev/hidraw0
+protocol=PI30
+command=QPIRI
+outputs=json_mqtt
+pause_loops=5
+mqtt_output_topic=green_grid/ross_proto_01/
--- a/mppsolar/__init__.py
+++ b/mppsolar/__init__.py
@@ -212,10 +212,10 @@ def main():
     if args.daemon:
         import time
 
-        import systemd.daemon
+        from systemd.daemon import notify, Notification
 
         # Tell systemd that our service is ready
-        systemd.daemon.notify("READY=1")
+        notify(Notification.READY)
         print("Service Initializing ...")
         # set some default-defaults
         pause = 60
@@ -259,6 +259,7 @@ def main():
             porttype = config[section].get("porttype", fallback=None)
             filter = config[section].get("filter", fallback=None)
             excl_filter = config[section].get("exclfilter", fallback=None)
+            pause_loops = config[section].get("pause_loops", fallback=0)
             #
             device_class = get_device_class(_type)
             log.debug(f"device_class {device_class}")
@@ -274,6 +275,7 @@ def main():
                 # mqtt_port=mqtt_port,
                 # mqtt_user=mqtt_user,
                 # mqtt_pass=mqtt_pass,
+                pause_loops=pause_loops,
             )
             # build array of commands
             commands = _command.split(",")
@@ -359,7 +361,7 @@ def main():
             # for item in mppUtilArray:
             # Tell systemd watchdog we are still alive
             if args.daemon:
-                systemd.daemon.notify("WATCHDOG=1")
+                notify(Notification.WATCHDOG)
                 print(
                     f"Getting results from device: {_device} for command: {_command}, tag: {_tag}, outputs: {_outputs}"
                 )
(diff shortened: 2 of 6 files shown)


rossandrews:
I have created a small ExpressJs project that demos how these changes can be used to expose commands to the mpp-solar service as an API. https://github.com/rossandrews/mpp-solar-api

jblance:
Wow - first off thanks for taking the time to contribute

I'm in 2 minds about this particular change
While I agree these functions are needed, the mpp-solar command (__init__ code) has gotten too complicated in my mind, so I was looking at fixing and incorporating this type of functionality in the 'new' `powermon` command
- let me think on if we can just merge this or whether it is best moved to powermon

Specific comments (I'm away from my dev setup atm):
- does this work for situations where no mqtt broker is defined?
- I was thinking of (and prefer) timings instead of loops as the 'spacer', so adding a repeat time type of config item only running the comand after that many seconds have passed (not sleeping like before)
- looks like good improvements on the systemd code - though someone was wanting support for rc/init systems too
- still has the big problem of too many device objects and too many connects and disconnects (inherited from the mpp-solar command complexity mentioned above)

Happy to discuss the above - though I maintain this I dont pretend to be right all the time (sometimes any of the time...)

rossandrews:
Thanks for maintaining this library, it's been awesome being able to get data and send commands to the mpp-solar hardware. It also introduced me to the MQTT protocol which I've been using to get data from a DIY ampmeter.

I am not familiar with powermon but any refactoring you want to do to simplify the init sounds good to me.

1. I have not tested it with no mqtt broker defined. Should I default it to localhost?
2. I like timings better as well. I just found my changes getting rather large and wanted to put in a pull request to start the discussion. For the timings I can see setting a time for the main loop to run and the commands can have separate timing that has to be a multiple of the main loop. We can default the main loop timing to 1 second.
3. The systemd changes were necessary to get it working, I'm glad they look good since I was winging it.
4. How this library talks with the hardware is still magic to me. To reduce the number of devices we can split commands and devices. Device would hold port and protocol. Commands would hold output format, output topic, and timing. Commands would also specific the device they should run on in the config. We would end up with one device holding multiple commands, could we reduce the connects and disconnects with that structure? I can draw up a diagram if that would help.

jblance:
> Thanks for maintaining this library, it's been awesome being able to get data and send commands to the mpp-solar hardware. It also introduced me to the MQTT protocol which I've been using to get data from a DIY ampmeter.

No problem - it was for my own use and then grew (a lot) from there
MQTT is definitely useful

> 
> I am not familiar with powermon but any refactoring you want to do to simplify the init sounds good to me.
> 
powermon is just a new command in the mpp-solar package - I'm using a new command so I dont break the existing usage of mpp-solar (and hence some of the concerns with the PR)

> 1. I have not tested it with no mqtt broker defined. Should I default it to localhost?

No - it also needs to work for people without MQTT

> 2. I like timings better as well. I just found my changes getting rather large and wanted to put in a pull request to start the discussion. For the timings I can see setting a time for the main loop to run and the commands can have separate timing that has to be a multiple of the main loop. We can default the main loop timing to 1 second.

I was thinking more of maintaining timings (a bit like you have started with runtime etc)

> 3. The systemd changes were necessary to get it working, I'm glad they look good since I was winging it.

Note: I havent tried them yet

> 4. How this library talks with the hardware is still magic to me. To reduce the number of devices we can split commands and devices. Device would hold port and protocol. Commands would hold output format, output topic, and timing. Commands would also specific the device they should run on in the config. We would end up with one device holding multiple commands, could we reduce the connects and disconnects with that structure? I can draw up a diagram if that would help.

Close to my thinking - this is what I am working on in the powermon command (powermon.py in same location at main __init__.py )  
this is subject to change as i try things and change my mind (hence havent really 'released' this command, though it is present to try)
- powermon instance only does one device, so run multiple instances/services if have multiple devices
- a device per actual hardware device (or connection to a hardware device)
- connect and disconnect to be moved out of the loop, so connect once and then loop (reconnecting with timeouts if needed)
- can have multiple commands, each with own delay/gap and each with own output(s)
- each output can be configured separately (filters tags topics etc)
- current yaml config example is below (and lives in powermon.py), expecting a config file to configure the run (possibly with command line overrides)

```yaml
  port:
    path: /dev/ttyUSB0
    type: test
    baud: 2400
  protocol: PI30
  mqttbroker:
    name: null
    port: 1883
    user: null
    pass: null
  adhoc_commands:
    topic: test/command_topic
    outputs:
      - name: screen
      - name: mqtt
  commands:
    - command: QPIGS
      gap: 10s
      outputs:
      - name: screen
      - name: mqtt
    - command: QPIRI
      gap: 1m
      outputs:
      - name: screen
        tag: testtag
        filter: volt
```

Something like that anyway....

rossandrews:
I had not looked into the powermon command until now. I see you already started on a very similar sort of change that I've been working on.I will close this request get caught up on the powermon command.

I agree with your general config structure. I was thinking of making the scheduled commands mutable from an API. That way the schedule can be adjusted without having to reboot the service. Maybe having the commands in a different JSON file would help enable that but I will think about it. I'll start some topics in the discussion section.

jblance:
Don't take what's in powermon already as decided, still very much working out the best way to go
Yeah I was also thinking about being able to update commands (and pause etc) while running. Got to balance flexibility with complexity tho

PR #215, opened 2022-04-23
Json UDP output

skoky:
Adding new output publishing JSON overt UDP broadcast packet. Adding new config praram `udpport` to set UDP broadcast publishing port. Tested with PI17 protocol on InfiniSolar 10k


jblance:
thanks heaps for you contribution

PR #267, opened 2022-10-14
Added QPIGS2 support for PIP4048MST with 3 PV

kasperbirk:
Added these to support MPP SOLAR inverter with 3 pv inputs such as pip4048mst

"response": [
            ["float", "PV2 Input Current", "A"],
            ["float", "PV2 Input Voltage", "V"],
            ["float", "Battery voltage from SCC 2", "V"],
            ["int", "PV2 Charging Power", "W"],
            ["string", "Device status", ""],
            ["float", "AC charging current", "A"],
            ["int", "AC charging power", "W"],
            ["float", "PV3 Input Current", "A"],
            ["float", "PV3 Input Voltage", "V"],
            ["float", "Battery voltage from SCC 3", "V"],
            ["int", "PV3 Charging Power", "W"],
            ["int", "PV total charging power", "W"],
        ],

From documentation:
image


jblance:
Does that inverter still respond with pi30 to a QPI command?

kasperbirk:
Hi
I am using qpigs and qpigs2 right now and getting all values correct. It’s a great peace of software you have made. I did choose it over another because of the high update frequency. 

Is it possible to send commands to the inverter via mqtt? Like POP, PCP and the other commands to change the settings?

jblance:
the reason i was asking about the QPI output is I am hoping that the devs bumped the protocol version - if i make the above changes to pi30max it will create errors for the other inverters that use this. if QPI has been bumped then we can use that for this, otherwise it will need to go to a version like pi30mst or similar (i wish they would change the protocol number when they change the protocol....)

kasperbirk:
Ahh ok i now I see the problem. It should work with a specific MST protocol 👍🏻

jblance:
Sorry, can you post the output of a `QPI` command against the inverter please

jblance:
can you rename your file to `pi30mst.py` and i can merge this and add the protocol

frederickjh:
@kasperbirk  (in case you missed this)
> can you rename your file to `pi30mst.py` and i can merge this and add the protocol

jblance (approved):

PR #276, opened 2022-12-12
Add support for Easun SV serie

sand1812:
Added support for Easun SV inverters.
Has been tested on SV-IV 5.6 with parallel boards.

All read requests works.
Some set requests return NACK
Some set requests not yet implemented (waiting to find out why some don't work)

### DIFF
--- a/mppsolar/protocols/pi18sv.py
+++ b/mppsolar/protocols/pi18sv.py
@@ -1,0 +1,1014 @@
+import logging
+
+from .abstractprotocol import AbstractProtocol
+
+from .protocol_helpers import crcPI as crc
+
+# from .pi30 import COMMANDS
+
+log = logging.getLogger("18sv")
+
+COMMANDS = {
+    "PI": {
+        "name": "PI",
+        "prefix": "^P005",
+        "crc": False,
+        "description": "Device Protocol Version inquiry",
+        "help": " -- queries the device protocol version",
+        "type": "QUERY",
+        "response": [["string", "Protocol Version", ""]],
+        "test_responses": [
+            b"^D00518;\x03\r",
+        ],
+    },
+    "T": {
+        "name": "T",
+        "prefix": "^P004",
+        "crc": False,        
+        "description": "Query current time",
+        "help": " -- queries current time from the Inverter",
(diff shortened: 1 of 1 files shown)


mppsolar/protocols/pi18sv.py
```diff
@@ -0,0 +1,1008 @@
...
+
+class pi18sv(AbstractProtocol):
+    def __str__(self):
+        return "PI18-SV protocol handler"
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__()
+        self._protocol_id = b"PI18SV"
+        self.COMMANDS = COMMANDS
+        self.STATUS_COMMANDS = ["PI","T","ET","EY","EM","ED","ID","VFW","PIRI","GS","MOD","FWS","FLAG","DI","CHGCR","MUCHGCR","PRI","PGS","ACCT","ACLT"]
```
jblance, line 881: STATUS_COMMANDS are the commands to run when a --getStatus command is issued. This list looks too long (normally it is 1 or 2 or 3 key commands)

mppsolar/protocols/pi18sv.py
```diff
@@ -0,0 +1,1008 @@
...
+        self.SETTINGS_COMMANDS = [
```
jblance, line 882: as above, this is the --getSettings command, so the list is likely too long

jblance (changes requested):
I've made a few comments, do they make sense?
Mostly the requested changes are to make the code easy to keep up to date for multiple different systems.

Thoughts?

sand1812:
Hi,

Your comments looks right. I've lot of work today & tomorrow. I’ll update my code later this week. But feel free to do it if you want !

Do you have any idea why the SET commands are mostly not working ? 
For example, according to doc, « BUCD500,540 » should be ok, but the inverter returns 0 (failed)

I attached the doc from easun for reference.

Kind regards,

Greg.

jblance:
sorry i didnt see the doc
re commands not working, first check with debug to make sure the correct command is being sent - commas in particular are a bit half removed
the other thing ive seen is different padding for values, sometimes extra leading seros are needed

Issue #269, opened 2022-11-05
How to send data as devices/groups to HomeAssistant


Saentist:
Currently I use this command 
```bash
mpp-solar -p /dev/hidraw1 -n VII -o hass_mqtt,screen -c QPIGS#QDI#QFLAG#QPIRI \
-q 192.168.1.12 --mqttuser user --mqttpass pass --mqtttopic /PV/status -n SOLAR_INVERTER --daemon
```
But in home assistant i get bunch of sensors no group/source etc.
How to get them in groups?
```
VII
 -QPIGS
 +- entities
 -QDI
 +- entities
 -QFLAG
 +- entities
 -QPIRI
 +- entities
```
![image](https://user-images.githubusercontent.com/16975883/200137786-10dfa0d9-02d8-4169-8c68-5650d576422f.png)

jblance:
Im not sure what you are after.
Are you wanting to change how it looks in the configure screen?
If you know what needs to be sent to the broker I can look at changing the message

Saentist:
Lets imagine that we have multiple inverters
only difference is in entities name string.
It will be more usefully if grouped by source.

jblance:
Im still not clear what you mean by 'group'
When I was using 2 inverters, I was using the 'tag' option to provide
entity names that we easy to select by wildcard

Is this what you mean?

Saentist:
I try to use this software
https://github.com/lluiscab/MPQTT
(very limited support to inverters commands response and direct crash in my case)
but perfect organized sending of data
I see Inverter sensors as "Device with sensors" with is more organized, then bunch of sensors
![image](https://user-images.githubusercontent.com/16975883/200408092-9c0911ca-5d3b-4221-89f6-d4fa1335bb5b.png)

jblance:
Can you grab some of the discovery mqtt messages from mpqtt

Saentist:
> Can you grab some of the discovery mqtt messages from mpqtt

Is there any guide how to do it?

jblance:
In 'Settings' for MQTT go to Configure
Then you can listen to messages:
(my autoconfigure prefix us homeassistant)
Need to grab the config message related to mpqtt

[image: image.png]

Saentist:
[mqtt-1.json.txt](https://github.com/jblance/mpp-solar/files/9955512/mqtt-1.json.txt)
this is mqtt info

jblance:
ok, first attempt in version 0.13.7 which uses the name field
[...]

Issue #243, opened 2022-07-26
[FR] add undocumented commands for inverters


Saentist:
There is a lot of commands with are supported by inverter but **mpp-solar** do not know how to describe
ex:
```
1.
Command: QVFW - Main CPU firmware version inquiry
------------------------------------------------------------
Parameter                       Value           Unit
main_cpu_firmware_version       VERFW:00041.15

2.
QVFW2
Parameter                       Value           Unit
error                           NAK

3.
Command: QVFW3 - No description found
------------------------------------------------------------
Parameter                       Value           Unit
warning                         No definition for command QVFW3 in protocol b'PI30'
response                        (VERFW:00002.40a

4.
QVFW4
Parameter                       Value           Unit
error                           NAK

5.
Command: QMN - No description found
------------------------------------------------------------
Parameter                       Value           Unit
warning                         No definition for command QMN in protocol b'PI30'
response                        (VMIII-5000¨'  <<<<Inverter Type

6.
Command: QBEQI - No description found
------------------------------------------------------------
Parameter                       Value           Unit
warning                         No definition for command QBEQI in protocol b'PI30'
response                        (0 060 030 010 030 58.40 000 120 0 0000HÞ

7.
Command: QMUCHGCR - Max Utility Charging Current Options inquiry
------------------------------------------------------------
Parameter                       Value           Unit
max_utility_charging_current    002             A
unknown_value_in_response_1     010
unknown_value_in_response_2     020
unknown_value_in_response_3     030
unknown_value_in_response_4     040
unknown_value_in_response_5     050
unknown_value_in_response_6     060
```
QVFW = OK
QVFW2, QVFW3, QVFW4, QMN, QBEQI , QMUCHGCR = KO

№ 2 and 4 don't have a formatting

other interesting QUERY commands
```
QMN (some product identification)
QSID (advanced version of QID)
QBEQI (battery charging parameters)
QET: Query total PV generated energy
QEYyyyy: Query PV generated energy of year
QEMyyyymm: Query PV generated energy of month
QEDyyyymmdd: Query PV generated energy of day
QLT: Query total output load energy
QLYyyyy: Query output load energy of year
QLMyyyymm: Query output load energy of monthQLMyyyymm: Query output load energy of month
QLDyyyymmdd: Query output load energy of dayQLDyyyymmdd: Query output load energy of day
```

jblance:
Try pi30max protocol

Saentist:
```
[...]

PR #296, opened 2023-01-08
Update hass_mqtt.py

michas79de:
Hi @jblance 

I'm using your great work for collecting inverter data into Home Assistant (HA), but I also like to improve it a bit:

1)
I think, if we use "binary_sensor" instead of "sensor" for boolean parameters in the mqtt topics, HA will discover them as binary sensors, too. We should then see a blue/yellow eye icon depending on state and 'state_color: true' in the entity card. [https://www.home-assistant.io/integrations/binary_sensor/] [https://www.home-assistant.io/integrations/mqtt/#mqtt-discovery]

2)
Parameters/Entities without unit don't need and better don't have "unit_of_measurement" declaration. Even with an empty "" unit statement HA uses them as with unit. So the history graph card in HA then shows an empty line graph instead of a horizontal bar chart. e.g.: https://imgur.com/3ha8K97 original graph and above a bar of a copy as a template entity without "unit_of_measurement" (sadly no change in the given time) [https://www.home-assistant.io/dashboards/history-graph]

I hope you can agree with my point of view, because I absolutely don't know how to change this in my running installation! And because I don't speak python you better doublecheck my changes! (There's no endif, isn't? Is it just the four spaces to right/left for loops?)

And if you take over my changes, I just have to enter "sudo pip install -U mppsolar" and
"systemctl --user restart mpp-solar" to update and restart mpp-solar on my pi?

Last but not least:
The PI30 command QMN queries the Inverter Model. This isn't included in your code and I didn't find the right file/place for a proposal. Maybe you find some time to add this. But while it's very clear to read (just with leading "(" and ending "³", it's no problem for me.

### DIFF
--- a/mppsolar/outputs/hass_mqtt.py
+++ b/mppsolar/outputs/hass_mqtt.py
@@ -52,23 +52,36 @@ class hass_mqtt(mqtt):
                 # <discovery_prefix>/<component>/[<node_id>/]<object_id>/config
                 # topic "homeassistant/binary_sensor/garden/config"
                 # msg '{"name": "garden", "device_class": "motion", "state_topic": "homeassistant/binary_sensor/garden/state", "unit_of_measurement": "°C"}'
-                topic = f"homeassistant/sensor/mpp_{tag}_{key}/config"
-                topic = topic.replace(" ", "_")
-                name = f"{tag} {_key}"
-                if unit == "W":
-                    payload = f'{{"name": "{name}", "state_topic": "homeassistant/sensor/mpp_{tag}_{key}/state", "unit_of_measurement": "{unit}", "unique_id": "mpp_{tag}_{key}", "state_class": "measurement", "device_class": "power", "force_update": "true" }}'
+                if unit == "bool":
+                    topic = f"homeassistant/binary_sensor/mpp_{tag}_{key}/config"
+                    topic = topic.replace(" ", "_")
+                    name = f"{tag} {_key}"
+                    payload = f'{{"name": "{name}", "state_topic": "homeassistant/binary_sensor/mpp_{tag}_{key}/state", "unique_id": "mpp_{tag}_{key}", "force_update": "true" }}'
+                    msg = {"topic": topic, "payload": payload}
+                    msgs.append(msg)
+                    topic = f"homeassistant/binary_sensor/mpp_{tag}_{key}/state"
+                    msg = {"topic": topic, "payload": value}
+                    msgs.append(msg)
                 else:
-                    payload = f'{{"name": "{name}", "state_topic": "homeassistant/sensor/mpp_{tag}_{key}/state", "unit_of_measurement": "{unit}", "unique_id": "mpp_{tag}_{key}", "force_update": "true" }}'
-                # msg = {"topic": topic, "payload": payload, "retain": True}
-                msg = {"topic": topic, "payload": payload}
-                msgs.append(msg)
-                #
-                # VALUE SETTING
-                #
-                # unit = data[key][1]
-                # 'tag'/status/total_output_active_power/value 1250
-                # 'tag'/status/total_output_active_power/unit W
-                topic = f"homeassistant/sensor/mpp_{tag}_{key}/state"
-                msg = {"topic": topic, "payload": value}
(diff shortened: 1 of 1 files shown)


jblance:
> Hi @jblance
> 
> I'm using your great work for collecting inverter data into Home Assistant (HA), but I also like to improve it a bit:
> 

thanks for the improvements
> 
> I think, if we use "binary_sensor" instead of "sensor" for boolean parameters in the mqtt topics, HA will discover them as binary sensors, too. We should then see a blue/yellow eye icon depending on state and 'state_color: true' in the entity card. [https://www.home-assistant.io/integrations/binary_sensor/] [https://www.home-assistant.io/integrations/mqtt/#mqtt-discovery]
> 
> 
> Parameters/Entities without unit don't need and better don't have "unit_of_measurement" declaration. Even with an empty "" unit statement HA uses them as with unit. So the history graph card in HA then shows an empty line graph instead of a horizontal bar chart. e.g.: https://imgur.com/3ha8K97 original graph and above a bar of a copy as a template entity without "unit_of_measurement" (sadly no change in the given time) [https://www.home-assistant.io/dashboards/history-graph]
> 

these sound good

> I hope you can agree with my point of view, because I absolutely don't know how to change this in my running installation! And because I don't speak python you better doublecheck my changes! (There's no endif, isn't? Is it just the four spaces to right/left for loops?)
> 

yes spacing is critical for python, loops are defined visually

> And if you take over my changes, I just have to enter "sudo pip install -U mppsolar" and "systemctl --user restart mpp-solar" to update and restart mpp-solar on my pi?
> 

you code is a little messed up, but I get what you mean, I will incorporate your changes directly and get a new version packaged

> Last but not least: The PI30 command QMN queries the Inverter Model. This isn't included in your code and I didn't find the right file/place for a proposal. Maybe you find some time to add this. But while it's very clear to read (just with leading "(" and ending "³", it's no problem for me.

oh, I'll get that added to PI30

michas79de:
> you code is a little messed up, but I get what you mean, I will incorporate your changes directly and get a new version packaged
> 
😎
¯\\_(ツ)_/¯
I think this red/green code compare thing messed it up (visual) more than it is. I just want the algorithm do

```
if boolean
    send as binary sensor w/o units declaration
else
    if power
        do the power thing
    else if unitless
        send w/o units declaration
    else
        send as usual
```

But of couse it's your's, feel free to make it like you want to! 👍

Thank you very much in advance!

jblance:
> 😎 ¯\_(ツ)_/¯ I think this red/green code compare thing messed it up (visual) more than it is. I just want the algorithm do
> 
> ```
> if boolean
>     send as binary sensor w/o units declaration
> else
>     if power
>         do the power thing
>     else if unitless
>         send w/o units declaration
>     else
>         send as usual
> ```
> 


Yeah, but its simpler to go 
```
if boolean
    send as binary sensor w/o units declaration
else if power
    do the power thing
else if unitless
    send w/o units declaration
else
    send as usual
```

and there is a few lines in the bool logic that are common, so can be pull out a level - dont take it as a criticism any help is appreciated, its just the code is complicated anyway so I try my best to make it as simple as possible (especially regard duplication) otherwise weird bugs can occur.

jblance:
er scratch all that, I see why you did it this way now - good job and ignore me

Saentist:
@michas79de 
What about `hassd_mqtt` module it's more organized in HA #269 

See also #243  `QPI#QMN#QGMN` is maybe a most accurate way to detect inverter.

michas79de:
> @michas79de What about `hassd_mqtt` module it's more organized in HA #269

I did'n knew that, when I startet with mpp-solar. But now I'm using a lot of the sensors in cards, graphs or other sensors in HomeAssistant. So I don't want to change them all again.
Also I have just this one device and I'm using several sections with unique tags for each command in the mpp-solar.conf. So there's no further need for more organization.
But thanks anyway!

Saentist:
Just change to `hassd_mqtt`, noting will change to sensors in HA, 
just will have more organized view in Mosquitto broker.
see images in #269

Sooner or later hass_mqtt will migrate to hassd_mqtt.

PR #299, opened 2023-01-11
Update hass_mqtt.py

michas79de:
Sadly Home Assistant only recognises binary sensor states, if they are `"ON"` or `"OFF"` (case sensitive). 🤦🏻‍♂️
So there's also the need of changing the values for export.
And also sadly the if condition differs `"0"` from `QPIWS` and `0` from `QPIGS`.
For the `QPIWS` warnings I cannot see an universal conditon. Maybe you can somehow use the command to include it also in the if.
(For me I added the tag from mpp-solar.conf section: `if unit == "bool" or value == "enabled" or value == "disabled" or tag == "myQPIWStag":`)


michas79de:
sorry, I made my comment after line 63 - should be after line 55

PR #310, opened 2023-02-06
HW Specs update

Saentist:
Not only developer boards support MPP-Solar


jblance:
thanks for the contribution

PR #312, opened 2023-02-23
Pi18 homeassistant

preussal:
Hi, 
I had some time again and was able to continue the PI18 protocoll.
Since I saw that you are already building with a home assistant, I made a few adjustments here. So that the energy data can also be evaluated by the inverter.

grafik
grafik
grafik


mppsolar/outputs/hassd_mqtt.py
```diff
@@ -131,10 +133,16 @@ def build_msgs(self, *args, **kwargs):
                 }
                 if device_class:
                     payload["device_class"] = device_class
-                if unit == "W":
-                    payload.update({"state_class": "measurement", "device_class": "power"})
+                #if state_class:
```
jblance, line 136: any reason you didnt go with this approach?

mppsolar/protocols/pi18.py
```diff
@@ -506,7 +610,8 @@ def __init__(self, *args, **kwargs) -> None:
             "MCHGV",
             "MUCHGC",
         ]
-        self.DEFAULT_COMMAND = "PI"
+        self.DEFAULT_COMMAND = ["PIRI"]
+        self.ID_COMMANDS = ["ID"]
```
jblance, line 614: this needs to be PI only, so getDeviceId is the same for all devices that use this protocol.

mppsolar/outputs/hassd_mqtt.py
preussal: I had the problem that my counter in the Home Assistant jumped from time to time, so I had commented on it.
Will test it again.

mppsolar/protocols/pi18.py
preussal: ahh ok, is put back on Pi

preussal:
At the KWH handover, I at the beginning with 
`"Device class": "Energy", "State_Class": "Total_increasing"`
but that didn't work. A 0 was handed over every 10-30 minutes and then I had created 14 mWh PV at the next calculation :)

In the end it only worked cleanly 
`"Device_Class": "Energy", "State_Class": "Total", "Last_Reset": Str (DateTime.now ())`
Since then no jumps

grafik

Except for the first entry in the Home Assistant, which of course is 14mWh
grafik

Small fun on the edge Home Assistant calculates the house consumption from the sources.
grafik
One Day 14 MWh WTF

jblance:
I think HomeAssistant energy monitor needs instant power readings (which it then adds up for kWh) not running energy totals.

frederickjh:
> In the end it only worked cleanly `"Device_Class": "Energy", "State_Class": "Total", "Last_Reset": Str (DateTime.now ())` Since then no jumps
> 
> grafik
> 
> Except for the first entry in the Home Assistant, which of course is 14mWh grafik

See the Home Assistant documentation [here](https://developers.home-assistant.io/docs/core/entity/sensor/#long-term-statistics) on [Long-term Statistics](https://developers.home-assistant.io/docs/core/entity/sensor/#long-term-statistics).

According to the documentation I believe the "Last_Reset" should be "not set or set to None". I believe this is the first example given below in the quoted text. 

I think that setting the "Last_Reset" to now might be what is causing the 14mWh spike at the beginning with Home Assistants built in logic.

> ### Entities representing a total amount

> Entities tracking a total amount have a value that may optionally reset periodically, like this month's energy consumption, today's energy production or the yearly growth of a stock portfolio. The sensor's value when the first statistics is compiled is used as the initial zero-point.
> How to choose state_class and last_reset
> 
> It's recommended to use state class total without last_reset whenever possible, state class total_increasing or total with last_reset should only be used when state class total without last_reset does not work for the sensor.
> 
> Examples
> 

>   -   The sensor's value never resets, e.g. a lifetime total energy consumption or production: state_class total, last_reset not set or set to None

>    -  The sensor's value may reset to 0, and its value can only increase: state class total_increasing. Examples: energy consumption aligned with a billing cycle, e.g. monthly, an energy meter resetting to 0 every time it's disconnected
>   -   The sensor's value may reset to 0, and its value can both increase and decrease: state class total, last_reset updated when the value resets. Examples: net energy consumption aligned with a billing cycle, e.g. monthly.
>  -   The sensor's state is reset with every state update, for example a sensor updating every minute with the energy consumption during the past minute: state class total, last_reset updated every state change.
> 
> #### State class total
> 
> For sensors with state class total, the last_reset attribute can optionally be set to gain manual control of meter cycles. The sensor's state when it's first added to Home Assistant is used as an initial zero-point. When last_reset changes, the zero-point will be set to 0. If last_reset is not set, the sensor's value when it was first added is used as the zero-point when calculating sum statistics.
> 
> To put it in another way: the logic when updating the statistics is to update the sum column with the difference between the current state and the previous state unless last_reset has been changed, in which case don't add anything.

Also see the examples that follow the quoted text.

frederickjh:
> I think HomeAssistant energy monitor needs instant power readings (which it then adds up for kWh) not running energy totals.

I have been thinking about this, and I am not sure I agree, if by "**instant power readings** you mean readings taken at regular intervals.

What is with the power in between the readings, how is it accounted for?

This is like a diabetic that takes a blood sugar level reading with a test strip vs. an embedded monitor that constantly checks the blood sugar level. With the first you only know the reading is accurate for that point in time. The second can show what it was over time.

Since the inverters already have "power meters" that track total PV generation and total load used and Home Assistant makes it possible to use these type of power meter readings (for more information on this please see the link I provided above to [Long-term Statistics](https://developers.home-assistant.io/docs/core/entity/sensor/#long-term-statistics)) to track usage through out the day, it make more sense to use them as they would be more accurate then reading at regular intervals that only show what it was at that point in time.

jblance:
I agree with your logic and some of the HA docs seem to agree but that's
not the way it seems to work (or maybe one of the class values is wrong)

Happy to make changes that work but haven't found the clear answer in the
HA docs

frederickjh:
HA Docs state for this use case:

`state_class` =  total
`last_reset` = not set or set to `None`

I am not certain that the `last_reset` in the commit above is correct for interfacing with HA. Docs do state that the sensors value can be manually reset but not with this type. In the commit `last_reset` is being set to the current date and time.


From the [link](https://developers.home-assistant.io/docs/core/entity/sensor/#long-term-statistics) above under: Long-term Statistics

----

#### Entities representing a total amount
Entities tracking a total amount have a value that may optionally reset periodically, like this month's energy consumption, today's energy production or the yearly growth of a stock portfolio. The sensor's value when the first statistics is compiled is used as the initial zero-point

##### How to choose state_class and last_reset

It's recommended to use state class total without last_reset whenever possible, state class total_increasing or total with last_reset should only be used when state class total without last_reset does not work for the sensor.

Examples

- The sensor's value never resets, e.g. a lifetime total energy consumption or production: **state_class total, last_reset not set or set to None**

----

I tried to test this development version of `mpp-solar` in a python environment. The HA energy panel saw the Total PV and Total Load but in core MQTT the new "TEST" inverter never showed up.

frederickjh:
> I tried to test this development version of `mpp-solar` in a python environment. The HA energy panel saw the Total PV and Total Load but in core MQTT the new "TEST" inverter never showed up.

I was mistaken. I thought a new device would be created but it put the test sensors under the working inverter device.

jblance:
> @jblance Here is a [commit](https://github.com/gadget78/inverter-homeassistant/commit/acffd74260dfe1d93962cc00a8ac75a5477df6d5) in another project where this was implemented that might be of help. Commit message says:
> 
> > Update mqtt-init.sh
> > added in sections to have ENERGY sections work in HA
> > it needs valid last_reset_topic , state_class, and device_class
> > but only for the ENERGY Wh specific readings, but improved the others too

Ive not completely gone through this, but it looks like last reset is a different topic...
```
 \"state_topic\": \""$MQTT_TOPIC"/sensor/"$MQTT_DEVICENAME"_"$MQTT_SERIAL"/$1\",
 \"last_reset_topic\": \""$MQTT_TOPIC"/sensor/"$MQTT_DEVICENAME"_"$MQTT_SERIAL"/$1/LastReset\",
```

But it may be only putting a date in the past
```
  mosquitto_pub \
        -h $MQTT_SERVER \
        -p $MQTT_PORT \
        -u "$MQTT_USERNAME" \
        -P "$MQTT_PASSWORD" \
        -i ""$MQTT_DEVICENAME"_"$MQTT_SERIAL"" \
        -t "$MQTT_TOPIC/sensor/"$MQTT_DEVICENAME"_"$MQTT_SERIAL"/$1/LastReset" \
		-r \
        -m "1970-01-01T00:00:00+00:00"
```

frederickjh:
Yes, it does look that way. I will try to setup and test locally what happens with LastRest unset or set to **None**.

PR #321, opened 2023-03-13
Added information about not using end of line comments and moved all …

### LINKED ISSUE #320: Comments in configuration file do not seem to be ignored properly
Not sure what is going on there exactly but it seems that mpp-solar does not ignore comments in the configuration file properly. 

I started out with the commented configuration file and modified it for my needs, but ran into a bunch of errors.. After removing both commented lines and end commented lines, it is now working.

Thanks for looking into this.

frederickjh:
…end of line comments above the lines they refer to and added empty lines to make it easier to read. Fixes #320

### DIFF
--- a/docs/configfile.md
+++ b/docs/configfile.md
@@ -3,28 +3,31 @@
 ```
 [SETUP]
 ### NOTE WELL: No end of line comments are supported!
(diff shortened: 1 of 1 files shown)


jblance:
thanks!!

PR #330, opened 2023-04-03
Proposed changes powermon to support adhoc commands and better scheduling

rossandrews:
I would like to get your input before starting on some changes to the powermon. This isn't a real pull request, I tried to write it up as a discussion but the code blocks weren't very good. I'm just getting familiar with yaml so consider this pseudo code.

I would like to separate the devices from the schedule of the commands they are going to run. I would also like to move the adhoc listener definition into a schedule. This will allow multiple devices to be able to listen for adhoc commands.

### DIFF
--- a/powermon.yaml
+++ b/powermon.yaml
@@ -55,10 +55,10 @@ mqttbroker:
   user: null
   pass: null
   adhoc_commands:
-  topic: Test_Inverter/commands
-  outputs:
-  - name: screen
-    tag: Test_Inverter
+    topic: Test_Inverter/commands
+    outputs:
+    - name: screen
+      tag: Test_Inverter
 daemon:
   type: systemd
   keepalive: 10
--- a/powermon.yaml
+++ b/powermon.yaml
@@ -1,6 +1,6 @@
 # yaml config for powermon
-devices:
-- name: Test_Inverter
+device:
+  name: Test_Inverter
   id: 123456789
   model: 8048MAX
   manufacturer: MPP-Solar
@@ -14,20 +14,7 @@ devices:
 scheduling:
   loop: 60
   schedules:
-  #Define a schedule for a device to listen to adhoc commands
-  - name: adhoc_listener
-    device_id: 123456789
-    type: loop
-    run_every_x_loops: 1 # runs every 60 seconds 
-    commands:
-    - command: adhoc #the command name could be anything
-      type: listen
-      command_topic: /mpp-solar/commands/ #mqtt topic that will listen for commands
-      outputs:
-      - name: mqtt
-        tag: Test_Inverter
   - name: QPIGS_5_minutes
-    device_id: 123456789
     type: loop
     run_every_x_loops: 5 #will run every 5 minutes
     commands:
@@ -42,7 +29,6 @@ scheduling:
       - name: screen
         tag: Test_Inverter
   - name: 3pm_check
-    device_id: 123456789
     type: time
     time: 15:00:00 #run at 3pm everyday
     commands:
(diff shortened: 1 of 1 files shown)


jblance:
My intention was/is to run an instance of powermon per device (mostly to limit complexity while giving flexibility). So the device is really the 'thing' that the yaml relates to

Adhoc commands use mqtt subscription at the moment which triggers a callback function (which effectively fires as soon as a message is sent)

What is the intention behind a schedule for the adhoc commands?

rossandrews:
One powermon instance per device makes sense. The point of a schedule for the adhoc command schedule was to allow for multiple devices to listen for commands. If it's one device per powermon instance then that is not necessary.

rossandrews:
Have you already started on the adhoc command work?

jblance:
yes, it kinda works, but only in a simple way.
ie send a command 'QPIGS' to the defined mqtt topic and the results go to a series of topics (with adhoc added to topic somewhere)
it feels like it needs more:
- have a simple and complex topic be able to be defined where complex allows config to be sent (ie result topic etc)
- some kinda logic the figures out if the message is a simple command or something more
- use simple commands only but have some better way to 'know' what to do with results...
- something else i havent thought of yet

rossandrews:
How about I make the adhoc commands a JSON message that contains extra configuration to run a complex command. Then I will change the main loop to run through a list of standardize commands that can come from the configuration or an adhoc command request. I'll start with complex commands then we can see if simple commands are necessary.

jblance:
> How about I make the adhoc commands a JSON message that contains extra configuration to run a complex command. Then I will change the main loop to run through a list of standardize commands that can come from the configuration or an adhoc command request. I'll start with complex commands then we can see if simple commands are necessary.

The main loop already caters for processing the adhoc commands if they are found - just need to change the processing logic from expecting a simple command, to expecting a JSON config and then processing it - obviously first thing to decide is 'what should the config be'

My thought would be to accept a full, single command stansa
````yaml
- command: QPIGS
    outputs:
    - name: screen
      format: raw
      #keep_case: True
      #remove_spaces: False
      #extra_info: True
      tag: Test_Inverter
      filter: ^ac|^pv
```

the problem with this is the onus on building the JSON is on the sending side (hence the hesitation)

jblance:
True. But adhoc should be a one off command whereas the others repeat

jblance:
I'm refactoring and don't want to overload the old classes with rhe new configs

jblance:
> We can create an API that connects with the broker and sends the message. That's what I did with this project https://github.com/rossandrews/mpp-solar-api, but it only worked with my old fork that heavily modified the mpp-solar command. Using docker compose I can add an API that starts at the same time and can build the command.

I wonder if it is worth having a lightweight web interface (not just an API), maybe just a place to submit a command in the first instance....

Saentist:
> > We can create an API that connects with the broker and sends the message. That's what I did with this project https://github.com/rossandrews/mpp-solar-api, but it only worked with my old fork that heavily modified the mpp-solar command. Using docker compose I can add an API that starts at the same time and can build the command.
> 
> I wonder if it is worth having a lightweight web interface (not just an API), maybe just a place to submit a command in the first instance....

let's forget about forking, and make merging ;)
https://github.com/lluisball/godenerg
https://github.com/prajna-pranab/converse
https://github.com/rossandrews/mpp-solar-api the missing remote command ;)

rossandrews:
I agree with the web interface idea. We could also change the scheduled commands on the fly rather than needing to change the configuration file and restart. First step is the API then we can build a web interface that calls it.

rossandrews:
I've created a new pull request with the code changes so I will close this one.

PR #331, opened 2023-04-08
Adhoc

rossandrews:
Made some large changes. Now the development docker compose will deploy an API that can send commands and receive the results. This is a draft pull request to get feedback on the changes I am in the process of making to powermon

List of changes in no particular order:
- Brought in the mpp-solar-api code and included it in docker-compose
- Added usb port type for powermon, that's what I am using
- Add mqtt and mqtt_html types for powermon output
- Code refactoring:
  - Removed dynamic library imports and made them explicit, makes for more readable code. Still need to make this more readable.
  - Started replacing kwargs with named arguments whenever I encounter them, they make code very hard to follow
- Read all the yaml config upfront then use objects 
- Created a Schedule object that is responsible for coordinating the main loop and watching for adhoc commands

What's remaining in the next few weeks:
- Move to a consistent object factory pattern for creation of CommandSchedules and Outputs
  - classmethods like those in Schedule.py seem like a good option
- Clean up any remaining kwargs
- Add filters back in, they were adding a lot of complexity during refactoring

Future work:
- Make a more robust API layer, I used nodejs but if you have any preferences just let me know
- Put a front end on top of the API layer, might use vuejs but again let me know if there are any preferences
- Create an API that will give information about powermon, available commands, current schedule, etc
- It might make sense to separate powermon into it's own repo.


mppsolar/ports/__init__.py
```diff
@@ -19,27 +22,18 @@ class PortType(Enum):
...
     # return None if port type is not defined
-    if porttype is None:
+    if portType is None:
         return None
 
-    # transform porttype for module lookup
-    porttype_id = f"{porttype.lower()}port"
-    # Try to import the porttype module with the supplied name (may not exist)
-    try:
-        port_module = importlib.import_module("mppsolar.ports." + porttype_id, ".")
```
rossandrews, line 34: These dynamic library imports were tough to follow. Replaced with regular imports.

mppsolar/sender/mqtt_json.py
rossandrews, line 1: This doesn't work yet

jblance:
(On my phone so please excuse typos etc)
Some of these changes diverge a bit from my intentions.
The dynamic loading is so only the modules needed are loaded - eg someone added db outputs but if you don't need them you don't need to install the packages. 
Similarly with the kwargs, different modules needed different arguments but I wanted to make it easy to switch from one (say output) module to a different one - [though this maybe fixed with a better config object]
Also with the new outputs I'm trying to have a "sender" that is just the transport and a "format" that takes rhe data and formats it as the old approach ended up with duplication everywhere. So for example mqtt should send messages via a mqtt broker (and maybe expects a topic and a message). If you want to send json the json formatter gives json. Theory is could send same json somewhere else [thus thought is not complete]

mppsolar/ports/__init__.py
jblance: Means import all modules and there are large variations

rossandrews:
I did a bit of research on the importlib and it doesn't seem to be a recommended approach. Is there a performance issue with loading a module then not using it? I believe the code readability is far more important to avoid confusion between multiple developers.

I think we can use optional parameters to give the same flexibility as kwargs while making the code easier to read.

For the outputs I was just cramming things together until I got an output so I could get the api running. So the structure of the output should separate the sender from the formatter. That makes sense, I can add an OutputSender interface and an OutputFormat interface with different implementations for the various options. I'll change the config so it includes a sender type and a format type per output.

jblance:
True re readability. I for one didn't want to install the db modules. Also the current/old ble library needs to be compiled so it causes problems if it's listed as a dependency.

jblance:
I started moving powermon into a separate folder (as you did maybe it should go to the separate repo). Still a but of a mess but I'll try to commit what ielve done when I get up

jblance:
I was trying to stay python (for no real reason) so was thinking of bottle for the web work (though haven't really thought through this side). Was thinking of not relying on mqtt but allowing web interface to running daemon

rossandrews:
I'm a fan of decoupling systems so I would have powermon only interact with other programs via mqtt. Then we create a companion API that can send commands via the mqtt broker and receive responses. If anyone wants output other than mqtt then they can create a small app that listens to the mqtt events and does whatever they need from there.

It would make sense to run have all the web side run on python based web server. I want to focus on the APIs more than the front end but I could also add a simple admin front end to get the ball rolling.

I'm hoping that this structure helps simplify the code but allows for easy integration with home assistant.

jblance:
> I'm a fan of decoupling systems so I would have powermon only interact with other programs via mqtt. Then we create a companion API that can send commands via the mqtt broker and receive responses. If anyone wants output other than mqtt then they can create a small app that listens to the mqtt events and does whatever they need from there.

that makes sense - happy to go this way



> It would make sense to run have all the web side run on python based web server. I want to focus on the APIs more than the front end but I could also add a simple admin front end to get the ball rolling.

I think one of the key functions of adhoc commands is changing settings:
[first] a simple web frontend that can send a command and display the result 
[next] a display of the settings (maybe showing defaults) and possibly and interface to change them (this as a precursor to HA integration of control not just display)

note some/many have multiple devices - ie I have inverters, BMS and a Shunt that all use the same code base

rossandrews:
> note some/many have multiple devices - ie I have inverters, BMS and a Shunt that all use the same code base
So one instance of powermon per device then a web frontend that can send requests to multiple powermon instances.

jblance:
> > note some/many have multiple devices - ie I have inverters, BMS and a Shunt that all use the same code base
> 
> So one instance of powermon per device then a web frontend that can send requests to multiple powermon instances.

yeah, using different topics for each instance

mppsolar/libs/schedule.py
```diff
@@ -0,0 +1,174 @@
...
+        self.commandType = commandType
+        self.outputs = outputs
+        self.port = port
+
+    def __str__(self):
+        return f"Command: {self.command}, CommandType: {self.commandType}, Outputs: {self.outputs}"
+    
+    def run(self):
+        log.debug(f"Running command: {self.command}")
+        self.port.connect()
```
jblance, line 169: this is a problem in a couple of porttype, ble doesnt like repeating connects/disconnects, victron shunt and ble continuously send data

mppsolar/libs/schedule.py
```diff
@@ -0,0 +1,174 @@
...
+        results = self.port.process_command(command=self.command)
+        for output in self.outputs:
+            log.debug(f"Output: {output}")
+            output.output(data=results)
```
jblance, line 173: this loses config info, the amount of which varies

mppsolar/libs/schedule.py
```diff
@@ -0,0 +1,174 @@
+from enum import StrEnum, auto
+from time import sleep, time
+import yaml
+import logging
+from mppsolar.sender import get_output
+
+log = logging.getLogger("schedule")
+
+class Schedule:
```
jblance, line 9: I like this concept but I think this misses a few things, especially for the more complicated porttypes
A lot of the complexity you are removing has been added to cover these other porttypes, a lot of the reason behind powermon and the expanded yaml config is to cover the variation between different hardware / porttypes

mppsolar/libs/schedule.py
rossandrews: USB required a new connect with every call. We will need to move the connection state into the device object so the different device types can handle it whichever way is more appropriate.

rossandrews:
I can't get docker to work with your change in structure. Did you try it out?

Docker compose file needs to be in the root directory to get access to all the files.

jblance:
> I can't get docker to work with your change in structure. Did you try it out?
> 
> Docker compose file needs to be in the root directory to get access to all the files.

No didnt try it yet, was short of time so was just breaking everything and then working through fixes - is the location related to the file or the directory you run the docker command from?

rossandrews:
I got it sorted on my fork. It was a tough merge. If you are going to make incremental changes can you do it in a development branch then do a pull request to master? Will give me a chance to see what you are working on.

mppsolar/version.py
```diff
@@ -1,5 +1 @@
-import configparser
-
-version = configparser.ConfigParser()
-version.read("pyproject.toml")
-__version__ = version["tool.poetry"]["version"].replace('"', "")
+__version__ = "0.15.30"
```
jblance, line 1: can you revert / change this to the configparser version

powermon/__init__.py
```diff
@@ -47,17 +43,17 @@ class ConfigError(Exception):
 """
 
 
-def readYamlFile(yamlFile=None):
-    _yaml = {}
-    if yamlFile is not None:
+def readConfigFile(configFile=None):
```
jblance, line 46: i renamed this intentionally

powermon/ports/__init__.py
```diff
@@ -1,8 +1,8 @@
 import importlib
 import logging
-from mppsolar.ports.serialport import SerialPort
-from mppsolar.ports.testport import TestPort
-from mppsolar.ports.usbport import USBPort
+from powermon.ports.serialport import SerialPort
+from powermon.ports.testport import TestPort
+from powermon.ports.usbport import USBPort
```
jblance, line 5: these are going to be problematic after the other ports are added, both from a memory perspective and a dependancy perspective
Ive been looking at poetry features to minimise dependancies so the modules are only added if the user needs them (especially, DB abd BLE modules)
Some users dont even use MQTT...

powermon/__init__.py
```diff
@@ -7,23 +7,19 @@
...
 
-from mppsolar.libs.schedule import Schedule, LoopCommandSchedule, CommandScheduleType, Command
+from powermon.libs.schedule import Schedule, LoopCommandSchedule, CommandScheduleType, Command
 
 # from mppsolar.inout import get_port
 
 
 # Set-up logger
-log = logging.getLogger("")
+log = logging.getLogger("Powermon")
```
jblance, line 22: I think this causes problems for logging from other files

mppsolar/version.py
rossandrews: It doesn't compile when I do.

powermon/__init__.py
rossandrews: You moved everything, it was a several hour merge effort. As a result some of your changes will be reverted.

mppsolar/version.py
jblance: how so? what fails?

rossandrews: mpp-solar-mppsolar-1      | Traceback (most recent call last):
mpp-solar-mppsolar-1      |   File "/usr/local/bin/powermon", line 5, in <module>
mpp-solar-mppsolar-1      |     from powermon import main
mpp-solar-mppsolar-1      |   File "/mpp-solar/powermon/__init__.py", line 10, in <module>
mpp-solar-mppsolar-1      |     from mppsolar.protocols import get_protocol
mpp-solar-mppsolar-1      |   File "/mpp-solar/mppsolar/__init__.py", line 5, in <module>
mpp-solar-mppsolar-1      |     from mppsolar.version import __version__  # noqa: F401
mpp-solar-mppsolar-1      |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
mpp-solar-mppsolar-1      |   File "/mpp-solar/mppsolar/version.py", line 5, in <module>
mpp-solar-mppsolar-1      |     __version__ = version["tool.poetry"]["version"].replace('"', "")
mpp-solar-mppsolar-1      |                   ~~~~~~~^^^^^^^^^^^^^^^
mpp-solar-mppsolar-1      |   File "/usr/local/lib/python3.11/configparser.py", line 979, in __getitem__
mpp-solar-mppsolar-1      |     raise KeyError(key)
mpp-solar-mppsolar-1      | KeyError: 'tool.poetry'
mpp-solar-mppsolar-1 exited with code 1

powermon/ports/__init__.py
jblance: there currently on 3 ports here and they use the same lib, mppsolar has 9(ish)
this import will fail if any dependant libs are missing, the current ble lib requires compiling
the old code did this and it caused a heap of trouble

mppsolar/version.py
jblance: bummer, so fails in the installed version
i really hated updating in 2 places

powermon/__init__.py
```diff
@@ -196,95 +189,45 @@ def main():
     #     protocol: PI30MAX
     port_config = device_config["port"].copy()
     log.debug(f"portconfig: {port_config}")
-    port = get_port(port_config)
+    portType = port_config["type"]
+    portPath = port_config["path"]
+    portBaud = port_config["baud"]
```
jblance, line 194: not all ports follow this config
this where i started and then ran into trouble with ever increasing parameters. this is why i have moved to passing the config object, so each port can have its own parameters

powermon/ports/__init__.py
```diff
@@ -19,27 +22,18 @@ class PortType(Enum):
...
-        port_class = getattr(port_module, porttype_id)
-    except AttributeError:
-        log.error(f"Module {port_module} has no attribute {porttype_id}")
-        return None
-    # Return the instantiated class
-    return port_class(config=config)
+    portObject = None
+
+    if portType == PortType.SERIAL:
+        portObject = SerialPort(portPath, baud, protocol)
```
jblance, line 33: if you want explicit loads, why not import the module here (then it doesnt get imported unless you want it)

powermon/ports/__init__.py
rossandrews: BLE is bluetooth low energy right? Is it used as a way to communicate with the hardware? I think we could move the importlib call to the specific port that causes problems. If that doesn't solve it we can revert back to dynamic imports.

I think DB output should be kept out of powermon and moved into a layer that listens to the MQTT events. Keep powermon focused on interacting with the hardware

jblance: > BLE is bluetooth low energy right? Is it used as a way to communicate with the hardware? I think we could move the importlib call to the specific port that causes problems. If that doesn't solve it we can revert back to dynamic imports.

yes. Some BMSs are currently supported via BLE and Im looking to extend the functionality to more BLE devices. I am hoping to move to a different library that doesnt need compiling (future milestone)



> I think DB output should be kept out of powermon and moved into a layer that listens to the MQTT events. Keep powermon focused on interacting with the hardware

I dont disagree at all (and dont use this myself) but some find it useful and have sent PRs (that have been accepted) for this type of feature
In my setup MQTT is on a different host (as my primary inverter location has intermittant network troubles), so I can (logically) see the use of local storage of data (or caching)

jblance:
Maybe its worth separating the elements of the PR - do one for docker and api as this seems straight forward

powermon/ports/__init__.py
rossandrews: Is anyone using powermon yet?

jblance:
Only me as far as I know

rossandrews:
Closing to open separate smaller pull requests.

PR #337, opened 2023-04-19
Adhoc

rossandrews:
There are a lot of changes, I hope you find they help improve the readability of the code without limiting flexibility. Next step is to create an API that allows for changes to the scheduled commands while powermon is running.

Some highlights
- objects are built from configuration and the imports are done in the if statement so only loaded if required. Configuration isn't passed to methods since objects already contain the state
- key formatting and filtering moved to the base class removing a lot of duplicate code. Take a look at htmltable
- Schedule listens for commands, currently only adds OneTimeCommandSchedules, but will be able to add other Schedule types in the future.

Next steps:
- Accept requests for information to output the details of what commands are scheduled
- Be able to modify and delete existing command schedules
- Create a web UI to send schedule changes to powermon

I'm going to be busy with work for the next few days but I'll start working on your feedback this weekend.


mppsolar/libs/mqttbrokerc.py
```diff
@@ -14,9 +14,18 @@ def __str__(self):
...
+
+        # mqttbroker:
+        #     name: localhost
+        #     port: 1883
+        #     user: null
+        #     pass: null
+        #     adhoc_commands:
+        #     topic: test/command_topic
+        self.config = config
+        if config is None:
```
jblance, line 27: Why change from {} as a default?

powermon/formats/abstractformat.py
```diff
@@ -0,0 +1,85 @@
...
+    @abstractmethod
+    def format(self, data):
+        pass
+
+    #Override this if the format sends multiple messages
+    def sendsMultipleMessages(self) -> bool:
+        return False
+
+    def formatAndFilterData(self, data):
+        #TODO: should we make data a proper object so it's easy to get the data we want?
```
jblance, line 43: Was just thinking this and yes I think we need a 'response' object (or similar) and embed the complexity there

powermon/ports/__init__.py
```diff
@@ -1,45 +1,37 @@
...
-    porttype = config.get("type", None)
+    if portType == PortType.SERIAL:
+        portObject = SerialPort(portPath, portBaud, protocol)
+    elif portType == PortType.USB:
+        portObject = USBPort(portPath, protocol)
 
-    # return None if port type is not defined
-    if porttype is None:
-        return None
+    #Pattern for port types that cause problems when imported
```
jblance, line 31: think we should do this for all ports - dont even need pyserial as a forced dependancy then

powermon/ports/serialport.py
```diff
@@ -2,52 +2,51 @@
...
 
     def disconnect(self) -> None:
         log.debug("usbserial port disconnecting")
-        if self.port is not None:
-            self.port.close()
+        if self.serialPort is not None:
+            self.serialPort.close()
         return
 
-    def send_and_receive(self, *args, **kwargs) -> dict:
```
jblance, line 35: wondering if this should use callbacks as the pattern
for normal serial not really a problem, but victron and daly and ble spew info continuously

powermon/powermon-qpigs.yaml
```diff
@@ -7,18 +7,25 @@ device:
...
\ No newline at end of file
+
+scheduling:
+  loop: once
+  schedules:
+  - name: QPIGS to screen
+    type: once
+    commands:
+    - command: QPIGS
+      type: basic #so far not neccessary, might remove
```
jblance, line 18: what did you envision as other options? (suspect that if you think there complicated use cases then you are probably correct - every time I have decided to assume only the simple case I have been bitten)

jblance:
Lots of change here, a few areas that I might disagree with (or need discussion) but overall looks good to me - I'm suggesting merging and dealing with any areas like above after merge (otherwise the delta gets too big)

mppsolar/libs/mqttbrokerc.py
rossandrews: Not an intentional change.

powermon/powermon-qpigs.yaml
rossandrews: Just seemed like a spot that would need some flexibility. Sounds like you need a different command type for devices that push data. A different command type will help differentiate the setup.

rossandrews:
> Lots of change here, a few areas that I might disagree with (or need discussion) but overall looks good to me - I'm suggesting merging and dealing with any areas like above after merge (otherwise the delta gets too big)
> 
> thoughts?

Sounds good to me. You could also make a new branch and I can merge there instead so you can try it out before it goes to master.

I'll mark this ready for review and leave it up to you.

PR #339, opened 2023-04-21
r0.16 pr

jblance:
not yet running / tested for change induced errors

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,2 @@
+[FORMAT]
+max-line-length=140
--- a/powermon/__init__.py
+++ b/powermon/__init__.py
@@ -1,9 +1,13 @@
 # !/usr/bin/python3
+"""main powermon code"""
+
 import logging
 from argparse import ArgumentParser
-from collections import deque
+
+# from collections import deque
 from datetime import date, timedelta  # noqa: F401
-from time import sleep, time
+
+# from time import sleep, time
 
 import yaml
 
@@ -15,7 +19,7 @@ from powermon.libs.mqttbroker import MqttBroker
 from powermon.libs.schedule import Schedule
 from powermon.libs.device import Device
 
-from powermon.ports import getPortFromConfig
+# from powermon.ports import getPortFromConfig
 
 
 # Set-up logger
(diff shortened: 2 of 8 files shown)


jblance:
@rossandrews first attempt at using a pr for discussion...

.pylintrc
jblance, line 1: i added pylint checking and this commit has a bunch of pylint changes (not sure if I like this or not, thoughts?)

powermon/__init__.py
```diff
@@ -125,13 +131,13 @@ def main():
         return None
 
     # Build configuration from defaults, config file and command line overrides
```
jblance, line 133: do we even need all this? maybe no sample_config, build objects from config file and change if overridden

powermon/__init__.py
```diff
@@ -142,89 +148,53 @@ def main():
         print("# default location ./powermon.yaml")
         print(yaml.dump(config))
         return
+    # debug config
+    log.info("config: %s", config)
 
-    # debug dump config
-    log.info(f"config: {config}")
+    # build mqtt broker object (optional)
+    mqtt_broker = MqttBroker(config=config.get("mqttbroker", {}))
```
jblance, line 155: think this is the best model, init the object with the config and handle errors, exceptions in the object class

powermon/__init__.py
```diff
@@ -142,89 +148,53 @@ def main():
...
     daemon.initialize()
-    
+
     # Main working loop
-    keepLooping = True
+    keep_looping = True
     try:
         schedule.beforeLoop()
-        while keepLooping:
+        while keep_looping:
```
jblance, line 186: while schedule.run_loop ?
or while true and schedule throws an exception when done??

.pylintrc
rossandrews: Adding lint is probably for the best. I've been naming variables with camelCase but that's the standard with Java. I should be following python standards and lint will help.

powermon/__init__.py
rossandrews: That makes sense.

powermon/__init__.py
rossandrews: I had it as while schedule.run_loop at first and you can change it back to that if it's more readable. I thought there might be a different condition that needs to change the loop in the future. I don't know what it would be yet though

jblance: We can also have exceptions (probably custom ones) to break the loop early and and use the catch blocks for dealing with these situations

PR #346, opened 2023-05-01
Release 0.16

jblance:


jblance:
@rossandrews attempt at merging your wip so things dont go too far apart...

rossandrews:
Thanks, I hope you like what you see. I'll be working on it again tomorrow and hopefully I'll get the device, schedule and results available through the api. I've been thinking about the structure of the objects in the dto folder quite a bit and have more changes.

jblance:
it is getting quite complicated - many users wont use the api so it needs to be optional

rossandrews:
It is deployed separately so it is optional. The api component could be moved into a new repo as long as it imports the DTO model from powermon to keep communication easy.

I was going to do a write-up for the pull request but I see you merged it to master. I'll add some thoughts here, we can move it to a discussion topic if you like.

- the API is implemented using FastAPI, it provides automated OpenAPI formatted documentation. This includes a simple UI to run commands
- I've introduced a shared data model being used by the API and Powermon in the dto folder. It's implemented using the pydantic library but could be replaced with the built in dataclass concept. It's still a work in progress
- The API is designed to work with multiple instance of powermon.
- Powermon announces itself on the MQTT broker so the API will automatically pick up new instances

In-progress:
- Clean-up the schedule part of the data model and how it's loaded from config
- Add all the endpoints to read the schedule information to the API
- Test and fix any issues I may have introduced with existing outputs and formats

next steps:
- Create new schedules from the API
- Add sample requests for the OpenAPI documentation
- Check that commands are part of the device protocol before creating them

jblance:
Discussion topic is a good idea
Im slowly going through the code and figuring out bits - I put some comments in the dynamic imports discussion

PR #353, opened 2023-05-09
Fastapi

rossandrews:
- Refactored the schedule classes into separate files that better match the existing patterns.
- Moved the DTOs into the Powermon folder since they are not used by mpp-solar
- Created a ConfigurationManager to handle the configuration parsing. Splitting it up was making it hard to work on
- Can get information about the powermon instances that have announced from the API.
- Changed the config as we discussed. Other configurations won't work until updated
- Moved all the yamls into a config folder, can be moved back it was just visual clutter
- Add an api_mqtt output

Planned Next steps:
- Remove the concept of loops and just use a duration
- Create all the API get end points
- Start on API post requests so modify powermon schedule and configuration


docker/dev/config/powermon.yaml
```diff
@@ -1,6 +1,8 @@
 # yaml config for powermon
```
jblance, line 1: i assume this is the 'current' working config file - think we need to delete the incorrect ones

powermon/libs/configurationManager.py
```diff
@@ -0,0 +1,150 @@
...
+    def parseControllerConfig(config : dict, device : Device, mqtt_broker : MqttBroker) -> ScheduleController:
+        logging.debug("parseCoordinatorConfig")
+        logging.debug(f"config: {config}")
+        _name = config["name"]
+        _loopDuration = config["loop_duration"]
+        if(_loopDuration == "once"):
+            _loopDuration = 0
+
+        _commands = []
+        topic_prefix = f"powermon/{_name}/results/"
```
jblance, line 122: this should be a config item

powermon/libs/configurationManager.py
```diff
@@ -0,0 +1,150 @@
...
+        if _commandType == CommandType.POLL:
+            _command = PollCommand(_command_query, _commandType, _schedule_name, _outputs, device.get_port())
+        
+        return _command
+    
+    @staticmethod
+    def parseControllerConfig(config : dict, device : Device, mqtt_broker : MqttBroker) -> ScheduleController:
+        logging.debug("parseCoordinatorConfig")
+        logging.debug(f"config: {config}")
+        _name = config["name"]
```
jblance, line 116: use config.get() instead after checking that we have a config - with sensible defaults or abort on None

powermon/libs/configurationManager.py
```diff
@@ -0,0 +1,150 @@
+
```
jblance, line 1: probally need something to check the config is valid

jblance:
I think the schedule part is too complicated, to my thinking there is really only
- a queue of commands that are either ready to run or are waiting till they are ready and
- an ability for commands to be run straight away

also functionality to look at and change these

powermon/libs/configurationManager.py
rossandrews: I've been a bit inconsistent with my config handling, I'll add get() here.

For configuration management, would you like to be able to run powermon without any configuration defined then fill it in through a web interface?

jblance: > 
> For configuration management, would you like to be able to run powermon without any configuration defined then fill it in through a web interface?

i think so
been trying to simplify things a bit too

powermon/libs/configurationManager.py
rossandrews: Yes, it needed to match with the API so I hardcoded it. I can include it in the announce message.

docker/dev/config/powermon.yaml
```diff
@@ -10,22 +12,22 @@ device:
...
-            type: htmltable
-            filter: ^ac|^pv
+schedules:
+- name: one_minute
+  type: loop
+  loop_count: 6 #will run every 1 minute
+
+commands:
+- name: QPIGS_every_minute
+  command_query: QPIGS #Must be in the protocol command list
```
jblance, line 22: why did you go to command_query?

rossandrews: the word command was really overused and I was getting them mixed up.

jblance: well thats true - think thats why i went for name
query implies something compared to the setters tho

rossandrews:
> I think the schedule part is too complicated, to my thinking there is really only
> 
>     * a queue of commands that are either ready to run or are waiting till they are ready and
> 
>     * an ability for commands to be run straight away
> 
> 
> also functionality to look at and change these

Adding APIs to get and modify the schedule is next on my list.

What did you have in mind in terms of simplifying? I am planning on adding a cron type schedule, should be useful to run commands on a daily basis or to run commands at a consistent interval so there isn't time drift that the loop will introduce.

It's still a work in progress so it may get more understandable as I continue making progress.

jblance:
i mean drop the schedules altogether and just 'schedule' the commands, eg
```yaml
commands:
- command: QPIGS
  type: basic #so far not neccessary, might remove
  trigger: # default to something sensible (only one of the below is needed/used)
  - loops: 6 # run every 6 loops
  trigger:
  - at: "15:00" # run at/near to 3pm localtime
  trigger:
  - every: 60 # run every 60 seconds (approx)
  trigger: ondemand # (maybe) to allow config to be setup in advance??
  trigger: disabled # excludes from 'once' runs
  outputs:
  - type: screen
```

or maybe without the trigger level and just
```yaml
commands:
- command: QPIGS
  type: basic #so far not neccessary, might remove
  - loops: 6 # run every 6 loops
  - at: "15:00" # run at/near to 3pm localtime
  - every: 60 # run every 60 seconds (approx)
  - ondemand # (maybe) to allow config to be setup in advance??
  - disabled # excludes from 'once' runs
  outputs:
  - type: screen
```

rossandrews:
That seems pretty similar, instead of a schedule containing many commands you will have a command containing a trigger. The code will be slightly more simple if it's a one-to-one relationship. If you want a command to have multiple triggers than it's the same code complexity as the current schedules.

jblance:
what is the use case for a command with multiple triggers?
it drops collections of schedules and just has commands (which are the _real_ components) - plus makes it simpler to have defaults and no config needed, eg
```yaml
commands:
- command: QPI
- command: QID
  trigger:
    at: "15:00"
```

jblance:
eg in branch using
```yaml
commands:
- command: QPIGS
  type: basic #so far not neccessary, might remove
  trigger:
    every: 25 # run every 25 seconds (approx)
  outputs:
  - type: screen
    format: simple
- command: QPI
- command: QID
  trigger:
    at: "12:56"
- command: QMON
  trigger:
    loops: 570
```

```
TODO: running command Command: QPIGS, type: basic, last run: 11 May 2023 12:55:26, next run: 11 May 2023 12:55:51, trigger: every, 25
TODO: running command Command: QPI, type: basic, last run: 11 May 2023 12:55:26, next run: 11 May 2023 12:56:26, trigger: every, 60
TODO: running command Command: QMON, type: basic, last run: 11 May 2023 12:55:26, next run: unknown, trigger: loops, 570, to go: 570
TODO: running command Command: QPIGS, type: basic, last run: 11 May 2023 12:55:51, next run: 11 May 2023 12:56:16, trigger: every, 25
TODO: running command Command: QID, type: basic, last run: 11 May 2023 12:56:00, next run: 12 May 2023 12:56:00, trigger: at, 12:56
TODO: running command Command: QPIGS, type: basic, last run: 11 May 2023 12:56:16, next run: 11 May 2023 12:56:41, trigger: every, 25
TODO: running command Command: QPI, type: basic, last run: 11 May 2023 12:56:26, next run: 11 May 2023 12:57:26, trigger: every, 60
TODO: running command Command: QPIGS, type: basic, last run: 11 May 2023 12:56:41, next run: 11 May 2023 12:57:06, trigger: every, 25
TODO: running command Command: QPIGS, type: basic, last run: 11 May 2023 12:57:06, next run: 11 May 2023 12:57:31, trigger: every, 25
```

rossandrews:
If you prefer that go for it. It simplifies the relationship so one command has one trigger. I am going to focus on getting the results from the API if you are going to work on changing main loop.

PR #356, opened 2023-05-23
V0.16d

jblance:


powermon/outputs/abstractoutput.py
```diff
@@ -1,22 +1,22 @@
...
     def __init__(self, formatter):
+        self.name = "AbstractOutput"
         self.formatter = formatter
 
     @abstractmethod
     def output(self, data):
         pass
+
+    @abstractmethod
+    def process(self, result):
```
rossandrews, line 21: It looks like this has the same purpose as the output method. I think we should just have one method to run the output.

jblance: yup, it just helps to transition tbh

powermon/outputs/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+    elif isinstance(outputConfig, list):
+        # eg [{'type': 'screen', 'format': 'simple'}], possibly multiple outputs
+        # loop through outputs
+        log.debug("got list type outputConfig: %s" % (outputConfig))
+        outputs = []
+        for item in outputConfig:
+            _out = parseOutputConfig(item, topic, schedule_name, device, mqtt_broker)
+            if _out is not None:
+                outputs.append(_out)
+        return outputs
```
rossandrews, line 108: A method shouldn't be able to return a single object or a list. It will make using it difficult. If we can return a list we should only return lists and set it as the return type.

jblance: true - good catch

powermon/commands/command.py
```diff
@@ -1,47 +1,58 @@
...
+            next_run = strftime("%d %b %Y %H:%M:%S", localtime(self.next_run))
 
-        return f"Command: {self.command}, type: {self.type}, last run: {last_run}, next run: {next_run}, {self.trigger}"
+        _outs = ""
+        for output in self.outputs:
+            _outs += str(output)
+
+        return f"Command: {self.name}, type: {self.type}, outputs: [{_outs}] last run: {last_run}, next run: {next_run}, {self.trigger}, defn: {self.command_defn}"
 
     def __init__(self, config):
```
rossandrews, line 28: I know you like the pattern of configuration parsing in the object init but I do find it very hard to follow. I think configuration parsing should be done outside of the object in a helper class, especially since I would like to create APIs that allow configuration to be changed and saved.

powermon/__init__.py
```diff
@@ -166,17 +154,29 @@ def main():
...
+    device_looping = True
     try:
         while keep_looping:
             # tell the daemon we're still working
             daemon.watchdog()
 
             # run schedule loop
-            keep_looping = controller.runLoop() or device.runLoop()
+            if controller_looping:
+                controller_looping = controller.runLoop()
```
rossandrews, line 166: If you don't like my schedule loop you can remove it. Now it's just broken because it's not properly calling the protocols properly.

powermon/commands/abstractCommand.py
```diff
@@ -6,22 +6,24 @@
 
 log = logging.getLogger("Command")
 
+
 class CommandType(LowercaseStrEnum):
     POLL = auto()
 
+
 class AbstractCommand(ABC):
```
rossandrews, line 14: Too many things called command, seeing as you are moving away from the schedule idea you can delete this class and stick with your new type of command.

powermon/commands/command.py
jblance: i dont see how its harder to follow - the object you are building has the logic on how to build
this doesnt stop subsequent changes 
i find having all this logic somewhere else counter intuative - objects should know about themselves, not some other thing

powermon/__init__.py
jblance: has been removed

powermon/commands/abstractCommand.py
jblance: have started to tidy up things

tried to get the api to run, but it was throwing an error before starting - do you have your start process? (I put in the wiki what steps I was taking)

rossandrews: In the docker-compose.development.yaml file it defines how to run the API. 

One thing that could be causing the issue is I am using docker with the compose plugin. Which is slightly different than the docker-compose standalone. I will switch to the docker-compose standalone since that's what you are already using.

I launch it with the makefile command you setup, docker-powermon-dev-up. You will notice that I removed the hyphen from that command since it wasn't working for me.

powermon/commands/command.py
rossandrews: I would argue that the configuration isn't part of the object, just the instructions on how to setup the object. Eventually these things will come from other sources, like API calls, that won't follow the same format as the yaml configuration. It also prevents us from using types on the init parameters to enforce the correct object types are being passed in.

The internal model should be one step removed from the configuration, API, or DB so that other patterns can be added as needed. Parsing the config in the constructor means I cannot reuse these constructors when I work on the API that will allow for adding commands and triggers to a live Powermon instance. If I want to add a new command to run from an API I'll have to go from JSON -> YAML -> Object.

jblance: hmm, there is sense in that
i like knowing where the logic is and having it close to the object though

jblance: what about this type of pattern?
```python
# build device object (required)
    device = Device()
    device.setupFromConfig(config=config.get("device"))
    device.addCommandsFromConfig(commandConfig=config.get("commands"))
```

rossandrews: I was trying out a pattern with the @staticmethod annotation for a method that parses the configuration and calls the constructor. I liked it because it keeps the configuration parsing in the class while also leaving the init alone. I ran into issues with building child objects like triggers. I'll do a bit of research and start a discussion about it.

rossandrews: > what about this type of pattern?
> 
> ```python
> # build device object (required)
>     device = Device()
>     device.setupFromConfig(config=config.get("device"))
>     device.addCommandsFromConfig(config=config.get("commands"))
> ```

I want the Device() to have types on the parameters. With a staticmethod it should look like:
`device = Device.setupFromConfig(config=config.get("device"))`

jblance: like this one you did?
why have the constructor with all the arguments? (as opposed to setting values later)
```python
@staticmethod
    def parseCommandConfig(command_config: dict, topic_prefix: str, mqtt_broker: MqttBroker, device: Device):
        _command_query = command_config.get("command_query")
        if not _command_query:
            log.info(f"No command_query for config: {command_config}")
            return None
        _commandType = command_config.get("type")
        _schedule_name = command_config.get("schedule_name")
        results_topic = topic_prefix + _schedule_name + "/results/" + _command_query

        _outputs = []
        for outputConfig in command_config.get("outputs"):
            _output = parseOutputConfig(outputConfig, results_topic, _schedule_name, device, mqtt_broker)
            logging.debug(f"output type: {_output}")
            _outputs.append(_output)

        _command = None
        if _commandType == CommandType.POLL:
            _command = PollCommand(
                _command_query,
                _commandType,
                _schedule_name,
                _outputs,
                device.get_port(),
            )

        return _command
```

rossandrews: Inside of Device.setupFromConfig you can pull all the values out of the config and pass it into the Device()

rossandrews: If you have all the information you need it's best to put it in the constructor, then you can have an immutable object. My background is in java which provides lots of benefits for immutable objects.

jblance: > > what about this type of pattern?
> > ```python
> > # build device object (required)
> >     device = Device()
> >     device.setupFromConfig(config=config.get("device"))
> >     device.addCommandsFromConfig(config=config.get("commands"))
> > ```
> 
> I want the Device() to have types on the parameters. With a staticmethod it should look like: `device = Device.setupFromConfig(config=config.get("device"))`

Doesnt allowing an empty (or light) constructor make other constructor approaches (eg api) easier?
eg creating (maybe not a device but who knows) via api and some field is invalid - instead of all failing you get an incomplete device (maybe with defaults in some places) that allows modification 'to get it correct'?

rossandrews: With your simplified command structure the staticmethod should look cleaner.

jblance: > If you have all the information you need it's best to put it in the constructor, then you can have an immutable object. My background is in java which provides lots of benefits for immutable objects.

Dont we want to be able to mutate objects though?

jblance: in some areas I think it might pay to separate things a bit more than I have done, what are your thoughts on something more like:
```python
    # build device object (required)
    device = Device.setupFromConfig(config=config.get("device"))
    log.debug(device)
    # add commands to device command list
    commandsConfig = config.get("commands")
    for commandConfig in commandsConfig:
        device.addCommandFromConfig(commandConfig)
```
(and do away with commandQueue which doesnt do much)

PS thanks for all your effort, input and discussions - they are really helpful

rossandrews: In our case, we probably want to be able to mutate our objects. However mutating an object in order to finish initializing it means you could potentially have an incomplete object. Also it means you have to open up more ways to mutate it than is probably necessary.

I'm used to working on a big dev team and being strict up front often saves time later. I'm probably being too picky, the thin constructor and a method to load from configuration will make it easier to my API work to continue.

jblance: > In our case, we probably want to be able to mutate our objects. However mutating an object in order to finish initializing it means you could potentially have an incomplete object. Also it means you have to open up more ways to mutate it than is probably necessary.

True, but we can try to make it complete even with little to no info supplied (but maybe useless) and allow changes to 'make it correct' - especially with the api etc


> I'm used to working on a big dev team and being strict up front often saves time later. I'm probably being too picky, the thin constructor and a method to load from configuration will make it easier to my API work to continue.

I can see how thats important - but I have found with this that 'we dont know what we'll need next' and so I have been trying to keep things flexible - mppsolar command got loads of extra command line options to allow for the next requirement and I dont really like where it ended. The reason I went (yaml) config file was to allow the configuration to be evolvable as needed.
eg mppsolar uses `-p test0` to specify which test response to return but with config i can add it as an option to the test port configuration
```yaml
  port:
    type: test
    response_number: 0
    protocol: PI30
```
similarly there are several 'serial' ports in mppsolar. one is just to change the data read from read_until('\n') to read until no more data - these should be able to be covered by one serial port codebase and configuration

rossandrews: > in some areas I think it might pay to separate things a bit more than I have done, what are your thoughts on something more like:
> 
> ```python
>     # build device object (required)
>     device = Device.setupFromConfig(config=config.get("device"))
>     log.debug(device)
>     # add commands to device command list
>     commandsConfig = config.get("commands")
>     for commandConfig in commandsConfig:
>         device.addCommandFromConfig(commandConfig)
> ```
> 
> (and do away with commandQueue which doesnt do much)
> 
> PS thanks for all your effort, input and discussions - they are really helpful

This looks good and what I was going to suggest. Also for the pattern you can use the classmethod annotation. https://stackoverflow.com/questions/136097/difference-between-staticmethod-and-classmethod

No problem with the help, this is a really valuable project and you've done a great job maintaining it.

PR #366, opened 2023-07-03
jkv11: Use proper battery_power precision.

proxer05:


jblance:
thanks for the contribution!

PR #378, opened 2023-07-26
Update prom.py

cordelster:
Make Prometheus output always have a prefix of 'mpp_solar' 
* Make output for strings always 1 
* # BREAKING from previous version, first change should keep that from happening again going forward.

### DIFF
--- a/mppsolar/outputs/prom.py
+++ b/mppsolar/outputs/prom.py
@@ -76,9 +76,10 @@ class prom(baseoutput):
         log.debug(f"displayData: {displayData}")
 
         # print data
+        print(f'machine_role{{role="mpp_solar"}} 1')
         for key in displayData:
             value = displayData[key][0]
             if isinstance(value, str):
-                print(f'{name}_{key}{{device="{dev}",cmd="{cmd}",myStr="{value}"}} 0')
+                print(f'mpp_solar_{key}{{inverter="{name}",device="{dev}",cmd="{cmd}",myStr="{value}"}} 1')
             else:
-                print(f'{name}_{key}{{device="{dev}",cmd="{cmd}"}} {value}')
+                print(f'mpp_solar_{key}{{inverter="{name}",device="{dev}",cmd="{cmd}"}} {value}')


jblance:
Can you give an expected output for this for a couple of simple commands

mppsolar/outputs/prom.py
```diff
@@ -76,9 +76,10 @@ def output(self, *args, **kwargs):
...
         # print data
+        print(f'machine_role{{role="mpp_solar"}} 1')
         for key in displayData:
             value = displayData[key][0]
             if isinstance(value, str):
-                print(f'{name}_{key}{{device="{dev}",cmd="{cmd}",myStr="{value}"}} 0')
+                print(f'mpp_solar_{key}{{inverter="{name}",device="{dev}",cmd="{cmd}",myStr="{value}"}} 1')
             else:
-                print(f'{name}_{key}{{device="{dev}",cmd="{cmd}"}} {value}')
+                print(f'mpp_solar_{key}{{inverter="{name}",device="{dev}",cmd="{cmd}"}} {value}')
```
jblance, line 85: is there supposed to be a `1` at the end of this line

why did you go for hardcoded mpp_solar components here and at line 79?

cordelster:
Yes will do.

mppsolar/outputs/prom.py
cordelster: Prometheus specifications require a value of 1 for what is string based values that exist.
The prefix is meant to be fixed for a particular purpose. Any "variables" in Prometheus should be as tags and why I moved the [name} with the tag of inverter. It's their standard and I'm just adhering to it. It also makes Dashboards in grafana easily portable as well.

Example as expected output from using -o prom:
```
mpp_solar_battery_charging_current{cmd="QPGS1",device="None",inverter="Inverter_1"} 0
mpp_solar_battery_charging_current{cmd="QPGS2",device="None",inverter="Inverter_2"} 1
mpp_solar_battery_charging_current{cmd="QPIGS",device="None",inverter="Inverter_1"} 0
mpp_solar_battery_charging_current{cmd="QPIGS",device="None",inverter="Inverter_2"} 1
```
Prefix mpp_solar should be constant tied to the value category in this segment battery_charginng_current so in essence I created a standard for use of mpp_solar to express categories to prometheus. All variables are then tags within that can be parsed as required. 

When node_exporter pulls in this file, it adds tags, instance="hosting_host_name" and job="node" and adds in the time segment.:

```
mpp_solar_battery_charging_current{cmd="QPGS1", device="None", instance="192.168.120.19:9100", inverter="Inverter_1", job="node"} | 0
mpp_solar_battery_charging_current{cmd="QPGS2", device="None", instance="192.168.120.19:9100", inverter="Inverter_2", job="node"} | 0
mpp_solar_battery_charging_current{cmd="QPIGS", device="None", instance="192.168.120.19:9100", inverter="Inverter_1", job="node"} | 0
mpp_solar_battery_charging_current{cmd="QPIGS", device="None", instance="192.168.120.19:9100", inverter="Inverter_2", job="node"}
```

I tossed this around a bit with the prometheus devs before I made and submitted the update, specially since it was a breaking change from what I had in my head in theory how to format things, though what I had originally submitted was broken as prometheus was designed. I had found some bad examples as well that supported my original theory how to form the output, though it didn't seem right with other data I have worked with. Having never made anything for prometheus ingest data and only a user of it, best I had was an itch that I had to contact the devs and ask...
This could also be expanded to add a group tag to identify a cluster of inverters working together from other clusters of inverters if say one node was collecting from 3 sets of 9 inverters each, though I expect the use case for that scenario is extremely low for even two sets. That would end up looking something like:
`mpp_solar_battery_charging_current{cmd="QPIGS", device="None",cluster="inverter group_1", instance="192.168.120.19:9100", inverter="Inverter_2", job="node"}`

Hope that makes since and helps clarify.

cordelster: I created a Discord channel in case is more useful: https://discord.gg/4Gj3hS4z

cordelster: Supporting document: https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/writing_exporters.md

PR #386, opened 2023-08-08
Add protocol for LV6548

cordelster:
Adjusted protocol from pi30max for MPP/Rich Solar 6548 where Charger Source Priority does nnot appear to follow the pi30 protocol of four types ["Utility first", "Solar first", "Solar + Utility", "Solar only"] and ignores "Utility first".


jblance (commented):
this should only have the differences from pi30 (or pi30max/mst) - check pi30max and pi30mst for examples
the goal is to not duplicate anything if possible

cordelster:
Done and submitted.

PR #391, opened 2023-08-15
Subscribe to command topics dynamically

rossandrews:
Still working on being able to see the config and then modify it with api calls. For now the commands that have triggers set to run with type api_mqtt will have a results topic that is automatically subscribed to when the announce is received.

More work to do on the results object so it's actually useful.


powermon/device.py
```diff
@@ -90,10 +97,12 @@ def runLoop(self, force=False):
             for command in self.commands:
                 if force or command.dueToRun():
                     # run command
-                    result = self.port.run_command(command)
+                    result: Result = self.port.run_command(command)
                     # decode result
-                    self.port.protocol.decode(result)
+                    self.port.get_protocol().decode(result)
```
jblance, line 102: whats the logic behind the change to get_protocol()?

powermon/commands/command.py
```diff
@@ -26,7 +28,7 @@ def __str__(self):
         return f"Command: {self.name=} {self.full_command=}, {self.type=}, [{_outs=}], {last_run=}, {next_run=}, {str(self.trigger)}, {self.command_defn=}"
 
     @classmethod
-    def fromConfig(cls, config=None):
+    def fromConfig(cls, config=None, mqtt_broker=None) -> "Command":
```
jblance, line 31: would really prefer fromConfig to be only config processing
passing mqtt object feels like a future problem (even tho i use similar in places), one problem in the mppsolar code is that it is hard to add extra/different approaches - eg when someone added DB storage a bunch of function definitions needed to extend to include the DB info/object
maybe device (or some more generic object) should wrap messaging etc - so device.publish() can do mqtt but could also do some other type of messaging....

powermon/commands/trigger.py
```diff
@@ -17,52 +18,52 @@ class TriggerType(LowercaseStrEnum):
 
 class Trigger:
     def __str__(self):
-        return f"trigger: {self.trigger} {self.value} loops togo: {self.togo}"
+        return f"trigger: {self.trigger_type} {self.value} loops togo: {self.togo}"
 
     @classmethod
     def fromConfig(cls, config=None):
         if not config:
```
jblance, line 25: Ive moved python requirement to >3.10, so can use switch approach ...

powermon/commands/command.py
rossandrews: Good point, it should only be config. Maybe a builder pattern can help here, the fromConfig can return a builder that we can use to add more information that doesn't come directly from the config.

powermon/commands/trigger.py
rossandrews: I am not sure what you mean. I just renamed this variable to help with readability.

powermon/device.py
rossandrews: By switching from direct variable access to a getter method we can add a return type. I added a python linter to my IDE and it's telling me all the python best practices which I am incorporating.

powermon/commands/trigger.py
jblance: ahh, was a comment related to the if, elif block - just a note as before I rejected strenum due to python version but now Ive been convinced to bump minimum version....;

rossandrews:
Removed the non configuration parameters in the from_config() method. Used setter methods later to set those values.

powermon/commands/command.py
rossandrews: I shouldn't be the one resolving it

powermon/device.py
jblance: > By switching from direct variable access to a getter method we can add a return type. I added a python linter to my IDE and it's telling me all the python best practices which I am incorporating.

Which linter? (and what IDE are you using?) - coz seems like sensible recommendations and would make sense for me to use it too

rossandrews: I'm using visual studio code with pylint

PR #401, opened 2023-09-11
Fix "Serial port not available" message in powermon for serial connection

oleks85:
There were two issues in powermon:
1. serial port was not initialised. Instead using abstract methods to connect and disconnect I decided to use "with" structure, it should close connection automatically
2. problem with haas formatter in mqtt output. Was not used argument in DTO class. 

PS: this tool is really great! Would like to help to improve it


powermon/commands/command.py
```diff
@@ -113,7 +113,6 @@ def to_dto(self):
         return CommandDTO(
             command_code = self.code,
             device_id=self.device_id,
-            result_topic = self.outputs[0].get_topic(),
```
rossandrews, line 116: The result topic is needed by the API to subscribe to command results. The abstract_output has a get_topic defined so other outputs shouldn't be impacted. What error caused you to remove this?

rossandrews:
These changes will cause the port to be connected and disconnected every time a command is sent, previously a port could be connected once then reused for future calls. I think I did something similar in the past and @jblance said it would cause issues with some types of ports, like bluetooth.

A different approach is adding a check to the main loop to connect the port if it's not already connected. Ports that don't need a persistent connection can do nothing and just connect during the calls. That used to be in the schedule_controller but when that was removed it got lost.

oleks85:
> These changes will cause the port to be connected and disconnected every time a command is sent, previously a port could be connected once then reused for future calls. I think I did something similar in the past and @jblance said it would cause issues with some types of ports, like bluetooth.
> 
> A different approach is adding a check to the main loop to connect the port if it's not already connected. Ports that don't need a persistent connection can do nothing and just connect during the calls. That used to be in the schedule_controller but when that was removed it got lost.

I just spotted that for usbport, port opens in send_and_receive method right before send command to inverter. So I assumed same approach should be for serial port as well, and it works :). There is no implementation for other ports like you mention for bluetooth so I was not aware of it.  

Any idea how to fix it properly and not break anything else? :)

rossandrews:
> I just spotted that for usbport, port opens in send_and_receive method right before send command to inverter. So I assumed same approach should be for serial port as well, and it works :). There is no implementation for other ports like you mention for bluetooth so I was not aware of it.
> 
> Any idea how to fix it properly and not break anything else? :)

Make calls to connect and disconnect in the main loop. The main loop should be pulled into a controller for easier testing as well.

oleks85:
> > I just spotted that for usbport, port opens in send_and_receive method right before send command to inverter. So I assumed same approach should be for serial port as well, and it works :). There is no implementation for other ports like you mention for bluetooth so I was not aware of it.
> > Any idea how to fix it properly and not break anything else? :)
> 
> Make calls to connect and disconnect in the main loop. The main loop should be pulled into a controller for easier testing as well.

It's even better. Thanks. I did't test usb port, there was logic to open port inside send method. I have changed it, hope did not broke anything.

powermon/commands/command.py
oleks85: The error was due to missed method in MQQT class

powermon/device.py
```diff
@@ -96,10 +96,13 @@ def runLoop(self, force=False) -> bool:
         runs them
         """
         time.sleep(0.1)
-        if self.commands is None:
+        if self.commands is None or len(self.commands) == 0:
             log.info("no commands in queue")
             return False
         else:
+            # open connection on port
+            self.port.connect()
```
oleks85, line 104: In debug mode I see that sometimes port open/closed without needs:
`2023-09-16 18:42:22,866:DEBUG:serialport:disconnect@51: usbserial port disconnecting
2023-09-16 18:42:22,966:DEBUG:serialport:connect@42: usbserial port connecting. path:/dev/moxa0, baud:2400
2023-09-16 18:42:22,966:DEBUG:serialport:disconnect@51: usbserial port disconnecting
2023-09-16 18:42:23,066:DEBUG:serialport:connect@42: usbserial port connecting. path:/dev/moxa0, baud:2400
2023-09-16 18:42:23,067:DEBUG:serialport:disconnect@51: usbserial port disconnecting
2023-09-16 18:42:23,167:DEBUG:serialport:connect@42: usbserial port connecting. path:/dev/moxa0, baud:2400
2023-09-16 18:42:23,167:DEBUG:serialport:disconnect@51: usbserial port disconnecting
2023-09-16 18:42:23,267:DEBUG:serialport:connect@42: usbserial port connecting. path:/dev/moxa0, baud:2400
2023-09-16 18:42:23,267:DEBUG:serialport:disconnect@51: usbserial port disconnecting
2023-09-16 18:42:23,368:DEBUG:serialport:connect@42: usbserial port connecting. path:/dev/moxa0, baud:2400
2023-09-16 18:42:23,368:DEBUG:device:runLoop@108: Running command: QPIGS`

I thought this is due to empty commands list but nope. It might be because `force or command.dueToRun()` is false. Would be great to open port only there are command that will be executed it will improve performance, as well as open port once to run all commands. 
Any thoughts on this?

rossandrews: The list of commands won't be empty when there is nothing to run. It will still have all the commands it's just that dueToRun() will return false.

I would put `self.port.connect()` in the `if force or command.dueToRun():` statement so it's only called if something is about to be run. This will cause it to be called for every command. To fix that, in the port.connect() add a check to see if it's already connected and do nothing if the connection is open.

powermon/device.py
```diff
@@ -96,10 +96,13 @@ def runLoop(self, force=False) -> bool:
...
+
             for command in self.commands:
                 if force or command.dueToRun():
                     log.debug(f"Running command: {command.code}")
```
rossandrews, line 108: Add `self.port.connect()` around here

powermon/device.py
rossandrews: Another option would be to add an a method to the AbstractPort called is_connected() that returns true if the port is already connected and check to see if it's already connected before trying. This is a bit more verbose but might make the main loop more readable. We also can also use it to enforce that each port implementation has that method so the code works as expected when we add more ports

rossandrews (commented):
There are some unit and integration tests in this project, there aren't many yet but it's still helpful to run. The command to run them is `make tests`

powermon/device.py
oleks85: What do you think if put connect and disconnect into initialize and finalize methods instead of checking if it connected or not in each command run?

rossandrews: That would leave a persistent connection which may not work for all ports. Calling it every command and managing the connection behaviour in the port implementation leaves more options open. @jblance would have a better idea about the needs of different ports. I only use USB, going to start using bluetooth as well in the next few months.

jblance: Bluetooth is particular bad with multiple reconnects - it causes the bluetooth stack to crash eventually (and on the JKBMS it beeps for every reconnect)
Victron VEDirect sends data continuously on serial as well
mpp-solar did the connect on each command and is fine for the simple case but doesnt work for many situations, hence the move to a maintained connection

jblance: I think it should be something like
- disconnect in finalize
- a connect for each command run, but with a is_connected test in the port.connect and so the connect runs if not connected but doent run if connected, the is_connected logic can then cover a bunch of cases
- possibly a `reconnect for each command` type config parameter that triggers a disconnect after the command run

I have a bad network at the inverter/battery shed and have been thinking about the need to deal with connection failures (not relevant here as much as for the mqtt connection, but similar logic may need to apply)

oleks85: What logic should be in `is_connected` method? Is it enough check if `serialPort` variable is not None? How can we check if connection is still alive?

powermon/device.py
```diff
@@ -96,21 +99,31 @@ def runLoop(self, force=False) -> bool:
...
-                    # decode result
-                    self.port.get_protocol().decode(result=result, command=command)
-                    result.set_device_id(self.device_id)
-                    # loop through each output and process result
-                    output: AbstractOutput
-                    for output in command.outputs:
-                        log.debug(f"Using Output: {output}")
-                        output.process(result=result)
+                    #check if port need to be connected
+                    if not self.port.isConnected():
```
jblance, line 109: add this check in the port, ie just do self.port.connect() here and then in the port itself check if anything needs to be done

powermon/device.py
```diff
@@ -96,21 +99,31 @@ def runLoop(self, force=False) -> bool:
...
+                        # open connection on port
+                        self.port.connect()
+
+                    # Run command if port is opened
+                    if self.port.isConnected():
```
jblance, line 114: actually, better still - run the command if it is supposed to run, deal with connect and isConnected etc in run_command
can then have disconnect login in run_command if wanted

powermon/ports/serialport.py
```diff
@@ -38,6 +38,9 @@ def toDTO(self) -> PortDTO:
         dto = PortDTO(type="serial", path=self.path, baud=self.baud, protocol=self.protocol.toDTO())
         return dto
 
+    def isConnected(self):
+        return self.serialPort is not None
```
jblance, line 42: maybe use pyserial `is_open` as well?

powermon/ports/usbport.py
```diff
@@ -48,16 +52,6 @@ def send_and_receive(self, command: Command) -> Result:
...
         result = Result(command_code=command.code)
 
-        # Open USB device
-        usb0 = None
-        try:
-            usb0 = os.open(self.path, os.O_RDWR | os.O_NONBLOCK)
-        except Exception as e:
-            log.error("USB open error: {}".format(e))
-            result.error = True
-            result.error_messages.append("USB open error: {}".format(e))
```
jblance, line 58: in connect() need a self.port = None for the exception path

tests/powermon/test_device.py
```diff
@@ -0,0 +1,67 @@
+from unittest import TestCase
+from unittest.mock import Mock
+from powermon import Device
+from powermon.ports import SerialPort
+from powermon.commands.command import Command
+from powermon.outputs.abstractoutput import AbstractOutput
+from powermon.protocols.pi30max import pi30max
+
+class DeviceTest(TestCase):
```
jblance, line 9: nice!!

tests/powermon/test_device.py
```diff
@@ -0,0 +1,67 @@
...
+                )
+            )
+        #Pretend that port is disconnected
+        self.port.isConnected.side_effect = [False, False]
+    
+        #Run main device loop. Expecting negative result due to closed port
+        self.assertFalse(self.device.runLoop())
+
+        #Check that port connection has been checked
+        self.port.isConnected.assert_called()
```
rossandrews, line 65: Mocks can be great but they add complexity to tests and make them harder to read and maintain. You have to be careful to test the code logic not the code execution path or your tests will make refactoring harder. If you find writing the unit tests are difficult without mocks it's a smell that the code has too many side effects and the code is highly coupled.

Instead of checking that isConnected() and connect() are called write a unit test to test the logic of those methods.

tests/powermon/test_serial_port.py
```diff
@@ -0,0 +1,23 @@
+from unittest import TestCase
+from unittest.mock import patch, Mock
+from powermon.ports.serialport import SerialPort
+
+class test_serial_port(TestCase):
```
rossandrews, line 5: Can you add a test that demonstrates whether a serial port keeps a persistent connection or reconnects with each call?

powermon/device.py
rossandrews: Yeah, the port can completely handle whether it's connected or not without other classes needing to check or take any action.

tests/powermon/test_device.py
rossandrews: A better use of mocks would be to check if output.process is called since that is the goal of the runLoop method.

oleks85: Mocks are very useful to simulate behaviour to properly test logic

rossandrews (approved):
I think this PR is good to go

rossandrews:
@jblance I would recommend you merge this PR in before you try to merge my huge refactor. I will merge this PR into my changes once you do.

jblance:
Sorry I'm on holiday so a bit slower than  normal..

PR #402, opened 2023-09-19
add QBMS command for pi30.py

lugamodder:
The QBMS command has been added according to the PI30 protocol description. It can be used to retrieve data from the BMS of a lithium battery connected to the inverter. Tested on Voltronic Axpert King with Pylontech US5000

### DIFF
--- a/mppsolar/protocols/pi30.py
+++ b/mppsolar/protocols/pi30.py
@@ -932,6 +932,56 @@ QUERY_COMMANDS = {
             b"",
         ],
     },
+    "QBMS": {
+        "name": "QBMS",
+        "description": "Read lithium battery information",
+        "help": " -- queries the value of various metrics from the battery",
+        "type": "QUERY",
+        "crctype": "chk",
+        "response": [
+
+            [
+                "keyed",
+                "Battery connect status",
+                {
+                    "0": "Connected",
+                    "1": "Disconnected",
+                },
+            ],
+            ["int", "Battery capacity from BMS", "%"],
+            [
+                "keyed",
+                "Battery force charging",
+                {
+                    "0": "No",
+                    "1": "Yes",
+                },
+            ],
+            [
+                "keyed",
+                "Battery stop discharge flag",
+                {
+                    "0": "Enable discharge",
+                    "1": "Disable discharge",
+                },
+            ],
+            [
+                "keyed",
+                "Battery stop charge flag",
+                {
+                    "0": "Enable charge",
+                    "1": "Disable charge",
+                },
+            ],
+            ["int", "Battery bulk charging voltage from BMS", "0.1 V"],
+            ["int", "Battery float charging voltage from BMS", "0.1 V"],
+            ["int", "Battery cut-off voltage from BMS", "0.1 V"],
+            ["float", "Battery max charging current", "A"],
+            ["float", "Battery max discharge current", "A"]],
+         "test_responses": [
+            b"(0 100 0 0 1 532 532 450 0000 0030\x5E\n",
+        ],
+    },
 }
 
 


jblance:
great thanks for the improvement
can you let me know what the result of `--getDeviceId` command is on your system, as pi30 has many versions (and the base/original version doesnt have QBMS, so I'd like to add it to a derivative version of pi30

lugamodder:
mppsolar -p /dev/ttyUSB0  -P pi30 --getDeviceId
Command: Get Device ID - Generate a device id
--------------------------------------------------------------------------------
Parameter Value          	Unit
deviceid  PI30:035:KING-5000	    
--------------------------------------------------------------------------------

mppsolar/protocols/pi30.py
```diff
@@ -932,6 +932,56 @@
...
+            ],
+            [
+                "keyed",
+                "Battery stop charge flag",
+                {
+                    "0": "Enable charge",
+                    "1": "Disable charge",
+                },
+            ],
+            ["int", "Battery bulk charging voltage from BMS", "0.1 V"],
```
jblance, line 976: these should be converted to Volts, ie something like `int:r*10`

jblance (changes requested):
can you migrate this to a separate protocol `pi30m035.py` much like [pi30m045.py](https://github.com/jblance/mpp-solar/blob/master/mppsolar/protocols/pi30m045.py) but just withg QMBS command added

jblance:
> > can you migrate this to a separate protocol `pi30m035.py` much like [pi30m045.py](https://github.com/jblance/mpp-solar/blob/master/mppsolar/protocols/pi30m045.py) but just withg QMBS command added
> 
> I am using PI30M045 currently, and moving the feature to an eventual PI30M035 protocol would be less convenient. If we do want it to be available for both xx35 and xx45, it would make sense to keep it in the _base_ PI30, WDYT (correct me if I'm wrong)?

The problem is not all devices with PI30 support that command (it would be great if the device manufacturers actually changed the protocol id when they changed the protocol commands)

> 
> For the record, I even have another device that works well with PI30M045, and would love to be able to see battery information too:
> 
> ```shell
> $ mpp-solar -P PI30M045 --getDeviceId
> ```
> 
> ```
> Command: Get Device ID - Generate a device id
> --------------------------------------------------------------------------------
> Parameter Value          	Unit
> deviceid  PI30:054:MKS2-5600	    
> --------------------------------------------------------------------------------
> ```

Yes there is a lot of commonality - but unfortuately also some differences (some significant) between different devices

> 
> Side note: I wanted to propose a wiki update for https://github.com/jblance/mpp-solar/wiki/Protocol-Recommendations. If you can do it, that would be great.

Whats the update?

BoboTiG:
> Whats the update?

1. In the Protocol Map section, add `PI30:054:MKS2-5600` with recommended protocol `PI30M045` and comment "Model 54 variant of PI30".
2. In the PI30 Model Details Table section, add `054 | MKS2-5600` at the bottom of the table.

jblance:
> > Whats the update?
> 
> 1. In the Protocol Map section, add `PI30:054:MKS2-5600` with recommended protocol `PI30M045` and comment "Model 54 variant of PI30".
> 2. In the PI30 Model Details Table section, add `054 | MKS2-5600` at the bottom of the table.

done

BoboTiG:
> The problem is not all devices with PI30 support that command (it would be great if the device manufacturers actually changed the protocol id when they changed the protocol commands)

Makes sense. Even documentations are messy :)

Anyway, lets do what you asked, and I will backport the command to PI30M045 if my tests are OK.

mppsolar/protocols/pi30.py
```diff
@@ -932,6 +932,56 @@
...
+            ["int", "Battery float charging voltage from BMS", "0.1 V"],
+            ["int", "Battery cut-off voltage from BMS", "0.1 V"],
```
BoboTiG, line 978: Tiny change, else the formatted parameter will be `battery_cut-off_voltage_from_bms`.
```suggestion
            ["int", "Battery cut off voltage from BMS", "0.1 V"],
```

PR #405, opened 2023-09-28
Refactoring formatting

rossandrews:
Formatting and responses are very connected so I've started making proper objects for each of the response definitions. By the end of this I'm hoping the AbstractProtocol class will be a lot smaller.

Still a work in-progress, will take me another week or two to get through it all. Starting this draft pull request to give a sense of how I'm refactoring this area. @jblance if you think I'm going off course let me know.


rossandrews:
All unit tests are now passing. Major refactoring to how results are processed and the abstract protocol is much smaller. This PR could be merged but there is still work to do.

- Move the python code in the protocol definitions into ResponseDefinition types, handle the information response types better, make a ResultType for commands that need info from the command being sent
- Make a Result type that handles the responses in different ways, move that logic out of the abstract protocol
- Have the protocol confirm the CRC codes are correct and strip those bytes before passing to the Result for processing into Response objects
- Write more unit tests for the refactored classes. Move the current tests that use subprocess into integration tests

powermon/config/config_model.py
```diff
@@ -30,7 +30,7 @@ class APIConfig(NoExtraBaseModel):
     """ model/allowed elements for api section of config """
     host: None | str
     port: None | int
-    enabled: None | bool
+    enabled: None | bool = Field(default=False)
```
jblance, line 33: what does this do?

powermon/config/config_model.py
```diff
@@ -127,7 +127,7 @@ class BaseConfig(NoExtraBaseModel):
     mqttbroker: None | MQTTConfig
     api: None | APIConfig
     daemon: None | DaemonConfig
-    debuglevel: None | str
+    debuglevel: None | int | str #If you put "debug" it translates to 10 then fails to load the config
```
jblance, line 130: can it be Logger.DEBUG ?

powermon/device.py
```diff
@@ -109,7 +109,14 @@ def runLoop(self, force=False) -> bool:
                     # run command
                     result: Result = self.port.run_command(command)
                     # decode result
-                    self.port.get_protocol().decode(result=result, command=command)
+                    try:
+                        self.port.get_protocol().decode(result=result, command=command)
+                    except Exception as e:
```
jblance, line 114: what are we catching here? can this not be part of decode?

powermon/formats/simple.py
```diff
@@ -20,29 +21,29 @@ def format(self, result: Result) -> list:
...
-            data = {}
-            data["Error"] = [f"Command: {result.command_code} incurred an error or errors during execution or processing", ""]
-            data["Error Count"] = [len(result.error_messages), ""]
-            for i, message in enumerate(result.error_messages):
-                data[f"Error #{i}"] = [message, ""]
-        else:
-            data = result.decoded_responses
-
-        if data is None:
+        #if result.error:
```
jblance, line 24: wont this mean errors dont get displayed?

powermon/protocols/pi30.py
```diff
@@ -11,7 +12,7 @@
         "name": "F",
         "description": "Set Device Output Frequency",
         "help": " -- examples: F50 (set output frequency to 50Hz) or F60 (set output frequency to 60Hz)",
-        "response_type": ResponseType.ACK,
+        "response_type": ResultType.ACK,
```
jblance, line 15: strictly this is a ack type response with a single result that is also an ack type

powermon/protocols/pi30.py
```diff
@@ -581,23 +582,23 @@
         "name": "QPI",
         "description": "Protocol ID inquiry",
         "help": " -- queries the device protocol ID. e.g. PI30 for HS series",
-        "response_type": ResponseType.INDEXED,
-        "response": [[0, "Protocol ID", "bytes.decode", ""]],
+        "response_type": ResultType.INDEXED,
+        "response": [[0, "protocol_id", "bytes.decode", ""]],
```
jblance, line 586: why the lower case, space removed? i dont think this is the place to do it

jblance (commented):
some great looking changes, im looking while away and on a laptop so havent had an opportunity to review in depth or run it up yet but have added a few comments and questions

powermon/commands/response.py
```diff
@@ -0,0 +1,31 @@
+from powermon.dto.response_dto import ResponseDTO
```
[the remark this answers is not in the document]
rossandrews: I've thought about that as well. At the moment the DTO is the only shared code the API needs to use which makes for a clean interface. I could be convinced otherwise but for now keeping them separate is clean.

powermon/protocols/pi30.py
rossandrews: There are two concepts here with the same name. I figured I would rename the higher level to a Result which contains different Responses. In this case the ResultType.ACK matches the ResponseType which is "ack". I will make a ResponseType enum and do a bulk replace to make it clear.

powermon/formats/simple.py
rossandrews: The errors are being added to the responses through exceptions at other places in the code.

powermon/commands/response.py
```diff
@@ -0,0 +1,31 @@
...
+        self.extra_info = extra_info
+        self.is_valid = True
+        
+    def to_DTO(self) -> ResponseDTO:
+        return ResponseDTO(data_name=self.get_data_name(), data_value=self.get_data_value(), data_unit=self.get_data_unit(), extra_info=self.get_extra_info())
+    
+    def __str__(self):
+        return f"Response: {self.data_name=}, {self.data_value=}, {self.data_unit=}, {self.extra_info=}"
+        
+    def get_data_name(self) -> str:
```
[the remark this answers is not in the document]
rossandrews: We can add a method if that's needed. Currently it's not used.

rossandrews:
> some great looking changes, im looking while away and on a laptop so havent had an opportunity to review in depth or run it up yet but have added a few comments and questions

Thanks for taking a look while away. It's no rush.

powermon/config/config_model.py
rossandrews: If the config isn't provided it sets the default to False. I suddenly found my config failing due to missing values. I figured sensible defaults could be added.

powermon/protocols/pi30.py
jblance: Both human format and data format (ie formated with spaces and all lower without spaces) are needed. Didn't really want to have separate fields for these, so had options for converting or not

powermon/protocols/pi30.py
jblance: My thinking was 'response' related to the response from the device, 'result' was the processed response, and '????' was each individual value/parameter

powermon/commands/response.py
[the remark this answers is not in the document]
jblance: api could just import the DTO class out of the same file as the main class
having them together makes keeping them in sync easier was my thinking

tests/powermon/unit/test_format_table.py
```diff
@@ -0,0 +1,70 @@
...
+                    'test       0.0     Check         ',]
+        table_formatter = table({})
+        table_formatter.draw_lines = False
+        
+        _result = Result(command_code=None)
+        response = Response(data_name="Test", data_value="0.0", data_unit="Check")
+        _result.add_responses([response])
+        
+        formatted_data = table_formatter.format(_result)
+        print(formatted_data)
```
jblance, line 41: need to comment out prints

tests/powermon/integration/__init__.py
jblance, line 1: seems like a bunch of failed tests in the integration suite?

powermon/commands/response_definition.py
```diff
@@ -74,63 +74,63 @@ def from_config(cls, response_definition_config : list) -> "ResponseDefinition":
...
-        elif response_definition_type == ResponseDefinitionType.FLAGS:
+        elif response_definition_type == ResponseType.FLAGS:
             flags : list[str] = response_definition_config[3]
             return ResponseDefinitionFlags(index=response_definition_index,
                                             description=response_definition_description,
                                             flags=flags,
                                             extra_info=response_definition_extra)
             
-        elif ResponseDefinitionType.INFO in response_definition_type:
+        elif ResponseType.INFO in response_definition_type:
```
jblance, line 154: how do we handle the template type responsetypes? ie where we have an int that needs to be *10 or /1000 or other re-calculations?

jblance:
thinking of commiting this - you happy for a merge at this stage?

tests/powermon/integration/__init__.py
rossandrews: Running 'make tests' or 'make powermon-integration-tests' I'm not seeing any errors

rossandrews:
> thinking of commiting this - you happy for a merge at this stage?

I think it's in a good state, merge away!

powermon/commands/response_definition.py
rossandrews: ResponseDefinitionInfo.translate_raw_response() uses the template. The template is pulled out on line 155 of response_definition.py

jblance: templates were not just for info responses, info was mostly 'parse part of the response into a new parameter'
there are also 'look at this other response based on this one' and 'the answer needs math applied' types are well

tests/powermon/integration/__init__.py
jblance: yeah, once merged i get no errors, the codespace against the PR showed errors though, weird

powermon/commands/response_definition.py
rossandrews: I didn't see any responses like that when looking around. If you have one in mind let me know where and I'll add a type for it.

jblance: they're not in protocols that have been ported yet, \
example from JK04
```
[
   "lookup:'Voltage Cell{:02d}'.format(m['Highest Cell'][0])",
   0,
   "Highest Cell Voltage",
   "V",
],
```
data has a 'Highest Cell' with a value that is a cell number and this looks up the voltage of that cell

another from pi17m058
```
[1, "Battery maximum charge current", "int:r/10", "A"],
```
result is a 0.1A answer, so we divide by 10 to get A, ie 240  0.1A is 24A


prolly dont need to be added yet, but just wanted you to be aware of stuff that will be needed

rossandrews: I would like to get rid of the code as strings and put them into methods for easier maintenance. We can add a whole bunch of response types. I'll work on simplifying how much code needs to go into each response type so adding new responses is easy. We can add INT_DIVIDE_BY_10 as a type in this case.

PR #410, opened 2023-10-15
Powermon PI18 first steps

preussal:
Hi, 
I've just made my first attempts at using the Powermon. And seen that only one protocol is stored.
I started writing the PI18 protocol

Here the first PIRI Output
```
2023-10-15 11:05:04,675:INFO:__init__:main@149: config: {'debuglevel': 10, 'device': {'name': 'Inverter1', 'id': 'a123456789', 'model': 'HybridV25048', 'manufacturer': 'MPP-Solar', 'port': {'path': '/dev/hidraw0', 'type': 'usb', 'protocol': 'PI18'}}, 'commands': [{'command': '^P007PIRI', 'type': 'basic', 'trigger': None, 'outputs': [{'type': 'screen', 'format': None}]}], 'loop': 'once', 'mqttbroker': None, 'api': {'host': '0.0.0.0', 'port': 5000, 'log_level': 'info', 'announce_topic': 'powermon/announce', 'adhoc_topic': 'powermon/adhoc', 'refresh_interval': 60}, 'daemon': None}
2023-10-15 11:05:04,676:DEBUG:mqttbroker:from_config@23: mqttbroker config: None
2023-10-15 11:05:04,676:INFO:__init__:main@154: MqttBroker DISABLED
2023-10-15 11:05:04,676:DEBUG:__init__:getPortFromConfig@12: port_config: {'path': '/dev/hidraw0', 'type': 'usb', 'protocol': 'PI18'}
2023-10-15 11:05:04,676:DEBUG:__init__:getPortFromConfig@21: portType: usb
2023-10-15 11:05:04,677:DEBUG:usbport:fromConfig@17: building usb port. config:{'path': '/dev/hidraw0', 'type': 'usb', 'protocol': 'PI18'}
2023-10-15 11:05:04,677:DEBUG:__init__:get_protocol_definition@13: Protocol PI18
2023-10-15 11:05:04,683:DEBUG:__init__:main@158: Device: Inverter1, self.device_id='a123456789', self.model='HybridV25048', self.manufacturer='MPP-Solar', port: <powermon.ports.usbport.USBPort object at 0x7ff116eaf140>, commands:[]
2023-10-15 11:05:04,683:DEBUG:__init__:getOutputs@48: processing outputsConfig: [{'type': 'screen', 'format': None}]
2023-10-15 11:05:04,683:DEBUG:__init__:parseOutputConfig@68: parseOutputConfig, config: {'type': 'screen', 'format': None}
2023-10-15 11:05:04,684:DEBUG:__init__:parseOutputConfig@90: got dict type outputConfig: {'type': 'screen', 'format': None}
2023-10-15 11:05:04,684:DEBUG:__init__:getFormatfromConfig@27: getFormatfromConfig, formatConfig: None
2023-10-15 11:05:04,684:DEBUG:__init__:getFormatfromConfig@39: getFormatfromConfig, formatType: simple
2023-10-15 11:05:04,685:DEBUG:__init__:parseOutputConfig@95: got format: Format: simple
2023-10-15 11:05:04,685:DEBUG:__init__:getOutputClass@23: outputType screen
2023-10-15 11:05:04,686:DEBUG:__init__:parseOutputConfig@97: got output: <powermon.outputs.screen.Screen object at 0x7ff116eaf230>
2023-10-15 11:05:04,687:DEBUG:command:__init__@29: Command: self.code='^P007PIRI' self.full_command=None, self.type='basic', [_outs='<powermon.outputs.screen.Screen object at 0x7ff116eaf230>'], last_run='Not yet run', next_run='15 Oct 2023 11:05:04', trigger: every 60 loops togo: 0, self.command_definition=None
2023-10-15 11:05:04,687:DEBUG:abstractprotocol:get_command_definition@102: Processing command '^P007PIRI'
2023-10-15 11:05:04,687:DEBUG:abstractprotocol:get_command_definition@104: Found command ^P007PIRI in protocol b'PI18'
2023-10-15 11:05:04,687:DEBUG:device:add_command@67: added command (Command: self.code='^P007PIRI' self.full_command=None, self.type='basic', [_outs='<powermon.outputs.screen.Screen object at 0x7ff116eaf230>'], last_run='Not yet run', next_run='15 Oct 2023 11:05:04', trigger: every 60 loops togo: 0, self.command_definition=<powermon.commands.command_definition.CommandDefinition object at 0x7ff116eaddc0>), command list length: 1
2023-10-15 11:05:04,688:INFO:__init__:main@165: Device: Inverter1, self.device_id='a123456789', self.model='HybridV25048', self.manufacturer='MPP-Solar', port: <powermon.ports.usbport.USBPort object at 0x7ff116eaf140>, commands:[<powermon.commands.command.Command object at 0x7ff116eaf260>]
2023-10-15 11:05:04,688:DEBUG:daemon:from_config@34: daemon config: None
2023-10-15 11:05:04,688:DEBUG:daemon:from_config@40: daemon not configured, disabling
2023-10-15 11:05:04,688:DEBUG:daemon:__init__@54: got daemon type: disabled, keepalive: 0
2023-10-15 11:05:04,689:INFO:__init__:main@169: Daemon DISABLED
2023-10-15 11:05:04,689:DEBUG:apicoordinator:from_config@23: ApiCoordinator config: {'host': '0.0.0.0', 'port': 5000, 'log_level': 'info', 'announce_topic': 'powermon/announce', 'adhoc_topic': 'powermon/adhoc', 'refresh_interval': 60}
2023-10-15 11:05:04,689:DEBUG:apicoordinator:set_mqtt_broker@56: MqttBroker DISABLED
2023-10-15 11:05:04,689:DEBUG:apicoordinator:set_mqtt_broker@57: No mqttbroker (or it is disabled) so disabling ApiCoordinator
2023-10-15 11:05:04,690:INFO:__init__:main@175: ApiCoordinator DISABLED
2023-10-15 11:05:04,690:INFO:device:initialize@90: initializing device
2023-10-15 11:05:04,790:DEBUG:device:run@107: Running command: ^P007PIRI
2023-10-15 11:05:04,791:DEBUG:usbport:connect@36: USBPort connecting. path:/dev/hidraw0, protocol: PI18 protocol handler
2023-10-15 11:05:04,849:DEBUG:usbport:connect@39: USBPort port number $3
2023-10-15 11:05:04,850:DEBUG:abstractport:run_command@65: Command Command: self.code='^P007PIRI' self.full_command=None, self.type='basic', [_outs='<powermon.outputs.screen.Screen object at 0x7ff116eaf230>'], last_run='Not yet run', next_run='15 Oct 2023 11:05:04', trigger: every 60 loops togo: 0, self.command_definition=<powermon.commands.command_definition.CommandDefinition object at 0x7ff116eaddc0>
2023-10-15 11:05:04,851:INFO:abstractprotocol:get_full_command@78: Using protocol b'PI18' with 36 commands
2023-10-15 11:05:04,851:DEBUG:protocol_helpers:crcPI@309: Calculating CRC for b'^P007PIRI'
2023-10-15 11:05:04,851:DEBUG:protocol_helpers:crcPI@359: Generated CRC 0xee 0x38 0xee38
2023-10-15 11:05:04,852:DEBUG:abstractprotocol:get_full_command@84: full command: b'^P007PIRI\xee8\r'
2023-10-15 11:05:04,852:DEBUG:result:__init__@30: Result: Result: self.is_valid=False, self.error=False - self.error_messages=[], self.raw_response=None, self.responses=[]
2023-10-15 11:05:04,853:DEBUG:usbport:send_and_receive@59: length of to_send: 12
2023-10-15 11:05:04,853:DEBUG:usbport:send_and_receive@68: multiple chunk send
2023-10-15 11:05:04,853:DEBUG:usbport:send_and_receive@75: sending chunk: b'^P007PIR'
2023-10-15 11:05:04,907:DEBUG:usbport:send_and_receive@75: sending chunk: b'I\xee8\r\x00\x00\x00\x00'
2023-10-15 11:05:07,013:DEBUG:usbport:send_and_receive@94: usb response was: b'^D0882300,217,2300,500,217,5000,5000,480,480,530,440,570,570,2,10,070,1,1,1,9,0,0,0,0,1,00\xe1k\r'
2023-10-15 11:05:07,015:DEBUG:abstractport:run_command@74: after send_and_receive Result: self.is_valid=False, self.error=False - self.error_messages=[], self.raw_response=b'^D0882300,217,2300,500,217,5000,5000,480,480,530,440,570,570,2,10,070,1,1,1,9,0,0,0,0,1,00\xe1k\r', self.responses=[]
2023-10-15 11:05:07,016:INFO:abstractprotocol:decode@146: result.raw_response passed to decode: b'^D0882300,217,2300,500,217,5000,5000,480,480,530,440,570,570,2,10,070,1,1,1,9,0,0,0,0,1,00\xe1k\r'
2023-10-15 11:05:07,017:DEBUG:protocol_helpers:crcPI@309: Calculating CRC for b'^D0882300,217,2300,500,217,5000,5000,480,480,530,440,570,570,2,10,070,1,1,1,9,0,0,0,0,1,00'
2023-10-15 11:05:07,019:DEBUG:protocol_helpers:crcPI@359: Generated CRC 0xe1 0x6b 0xe16b
2023-10-15 11:05:07,020:DEBUG:pi18:check_response_valid@821: CRCs match
2023-10-15 11:05:07,022:DEBUG:abstractprotocol:decode@169: trimmed and split responses: [<powermon.commands.response.Response object at 0x7ff116ecc110>, <powermon.commands.response.Response object at 0x7ff116ecc170>, <powermon.commands.response.Response object at 0x7ff116ecc1d0>, <powermon.commands.response.Response object at 0x7ff116ecc230>, <powermon.commands.response.Response object at 0x7ff116ecc2c0>, <powermon.commands.response.Response object at 0x7ff116ecc320>, <powermon.commands.response.Response object at 0x7ff116ecc380>, <powermon.commands.response.Response object at 0x7ff116ecc3e0>, <powermon.commands.response.Response object at 0x7ff116ecc440>, <powermon.commands.response.Response object at 0x7ff116ecc4a0>, <powermon.commands.response.Response object at 0x7ff116ecc500>, <powermon.commands.response.Response object at 0x7ff116ecc560>, <powermon.commands.response.Response object at 0x7ff116ecc5c0>, <powermon.commands.response.Response object at 0x7ff116ecc5f0>, <powermon.commands.response.Response object at 0x7ff116ecc650>, <powermon.commands.response.Response object at 0x7ff116ecc6b0>, <powermon.commands.response.Response object at 0x7ff116ecc6e0>, <powermon.commands.response.Response object at 0x7ff116ecc710>, <powermon.commands.response.Response object at 0x7ff116ecc740>, <powermon.commands.response.Response object at 0x7ff116ecc7d0>, <powermon.commands.response.Response object at 0x7ff116ecc800>, <powermon.commands.response.Response object at 0x7ff116ecc830>, <powermon.commands.response.Response object at 0x7ff116ecc860>, <powermon.commands.response.Response object at 0x7ff116ecc890>, <powermon.commands.response.Response object at 0x7ff116ecc8f0>, <powermon.commands.response.Response object at 0x7ff116ecc950>]
2023-10-15 11:05:07,023:DEBUG:device:run@123: Using Output: <powermon.outputs.screen.Screen object at 0x7ff116eaf230>
2023-10-15 11:05:07,024:INFO:screen:process@23: Using output sender: screen
2023-10-15 11:05:07,026:DEBUG:screen:process@24: formatter: Format: simple
2023-10-15 11:05:07,027:INFO:device:finalize@94: finalizing device
2023-10-15 11:05:07,028:DEBUG:usbport:disconnect@46: USBPort disconnecting 3
2023-10-15 11:05:07,030:DEBUG:mqttbroker:stop@114: Stopping mqttbroker connection
ac_input_voltage=230.0V
ac_input_current=21.7A
ac_output_voltage=230.0V
ac_output_frequency=50.0Hz
ac_output_current=21.7A
ac_output_apparent_power=5000VA
ac_output_active_power=5000W
battery_voltage=48.0V
battery_re-charge_voltage=48.0V
battery_re-discharge_voltage=53.0V
battery_under_voltage=44.0V
battery_bulk_charge_voltage=57.0V
battery_float_charge_voltage=57.0V
battery_type=User
max_ac_charging_current=10A
max_charging_current=70A
input_voltage_range=UPS
output_source_priority=Solar - Battery - Utility
charger_source_priority=Solar + Utility
max_parallel_units=9units
machine_type=Off Grid
topology=transformerless
output_mode=single machine output
solar_power_priority=Battery-Load-Utiliy + AC ChargerLoad-Battery-Utiliy
mppt_strings=1
unknown_flags?=00
```

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -53,3 +53,15 @@ services:
    devices:
      - /dev/ttyUSB0:/dev/ttyUSB0
 ```
+
+Build Dev Container
+change to Git Directory 
+```
+docker build -t YOUR_TAG -f docker/dev/Dockerfile .
+```
+
+Stop mpp-sor container and Run Powermon one time
+```
+docker compose -f /etc/docker/docker-compose.yaml stop; sleep 3; docker run --rm --volume=/etc/mpp-solar:/etc/mpp-solar --device=/dev/hidraw0 pi18_powermon powermon -C /etc/mpp-solar/powermon.conf -1; docker compose -f /etc/docker/docker-compose.yaml start;
+
+```
--- a/powermon/commands/response_definition.py
+++ b/powermon/commands/response_definition.py
@@ -8,6 +8,7 @@ import calendar #needed for INFO type evaluating templates
 class ResponseType(LowercaseStrEnum):
     ACK = auto()
     INT = auto()
+    TEN_INT = auto()
     OPTION = auto()
     BYTES = "bytes.decode" #can't use auto() for this value
     FLOAT = auto()
@@ -92,7 +93,16 @@ class ResponseDefinition(ABC):
                                          device_class=device_class,
                                          state_class=state_class,
                                          icon=icon)
-            
+        
+        elif response_definition_type == ResponseType.TEN_INT:
+            unit = response_definition_config[3]
+            return ResponseDefinitionTenInt(index=response_definition_index,
+                                         description=response_definition_description,
+                                         unit=unit,
+                                         device_class=device_class,
+                                         state_class=state_class,
+                                         icon=icon)
+        
         elif response_definition_type == ResponseType.OPTION:
             options : list[str] = response_definition_config[3]
             return ResponseDefinitionOption(index=response_definition_index,
@@ -227,6 +237,39 @@ class ResponseDefinitionInt(ResponseDefinition):
     
     def translate_raw_response(self, raw_value) -> str:
         return int(raw_value.decode())
+    
+    
+    def response_from_raw_values(self, raw_value) -> list[Response]:
+        value = self.translate_raw_response(raw_value)
+        return [Response(data_name=self.description,
+                        data_value=str(value),
+                        data_unit=self.unit,
+                        device_class=self.device_class,
+                        state_class=self.state_class,
+                        icon=self.icon)]
+    
+    def get_description(self) -> str:
+        return self.description
+      
+# Same INT but base 10
+class ResponseDefinitionTenInt(ResponseDefinition):
+    def __init__(self, index: int, description: str, unit : str,
+                 device_class: str = None,
+                 state_class: str = None,
+                 icon: str = None):
+        self.index = index
+        self.description = description
+        self.unit = unit
+        self.device_class = device_class
+        self.state_class = state_class
+        self.icon = icon
+    
+    def is_valid_response(self, value) -> bool:
+        #Do we need to check if it's negative? Should we have max and min bounds?
+        return isinstance(value, int) / 10
+    
+    def translate_raw_response(self, raw_value) -> str:
+        return int(raw_value.decode()) / 10
     
     
     def response_from_raw_values(self, raw_value) -> list[Response]:
--- a/docker/dev/config/powermon-qpigs.yaml
+++ b/docker/dev/config/powermon-qpigs.yaml
@@ -17,7 +17,7 @@ commands:
     # format: raw
     # format: simple
     # format: table
-    format: 
+    format:
       type: htmltable
       #keep_case: True
       #remove_spaces: False
--- a/docker/dev/config/powermon.yaml
+++ b/docker/dev/config/powermon.yaml
@@ -18,10 +18,9 @@ commands:
     every: 100  # seconds
   outputs:
   - type: screen
-    format: 
+    format:
       type: hass
 
-          
 mqttbroker:
   name: localhost
   port: 1883
--- a/powermon/protocols/pi18.py
+++ b/powermon/protocols/pi18.py
@@ -1,0 +1,842 @@
+import logging
+
+from powermon.commands.result import ResultType
+from powermon.protocols.abstractprotocol import AbstractProtocol
+from mppsolar.protocols.protocol_helpers import crcPI as crc
+from powermon.commands.result import Result
+from powermon.commands.response_definition import ResponseType
+
+log = logging.getLogger("pi18")
+
+SETTER_COMMANDS = {
+    "MCHGC": {
+        "name": "MCHGC",
+        "prefix": "^S013",
+        "description": "Set Battery Max Charging Current Solar + AC             (Manual Option 11)",
+        "help": " -- example: MCHGC0,030        (set unit 0 [0-9] to max charging current of  30A [    010 020 030 040 050 060 070 080])",
+        "response_type": ResultType.ACK,
+        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+        "regex": "MCHGC([0-9],0[1-8]0)$",
+    },
+    "MUCHGC": {
+        "name": "MUCHGC",
+        "prefix": "^S014",
+        "description": "Set Battery Max AC Charging Current                     (Manual Option 13)",
+        "help": " -- example: MUCHGC0,030       (set unit 0 [0-9] utility charging current to 30A [002 010 020 030 040 050 060 070 080])",
+        "response_type": ResultType.ACK,
+        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+        "test_responses": [b"",],
+        "regex": "MUCHGC([0-9]),(002|0[1-8]0)$",
+    },
+    "PBT": {
+        "name": "PBT",
+        "prefix": "^S007",
+        "description": "Set Battery Type                                        (Manual Option 14)",
+        "help": " -- example: PBT0              (set battery as PBT0 [0: AGM], PBT1 [1: FLOODED], PBT2 [2: USER])",
+        "response_type": ResultType.ACK,
+        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+        "regex": "PBT([012])$",
+    },
+    "PCP": {
+        "name": "PCP",
+        "prefix": "^S009",
+        "description": "Set charging source priority                            (Manual Option 10)",
+        "help": " -- example: PCP0,1            (set unit 0 [0-9] to 0: Solar first, 1: Solar and Utility, 2: Only solar)",
+        "response_type": ResultType.ACK,
+        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+        "test_responses": [b"^1\x0b\xc2\r", b"^0\x1b\xe3\r",],
+        "regex": "PCP([0-9],[012])$",
+    },
+#    "PE": {
+#        "name": "PE",
+#        "description": "Set the enabled state of an Inverter setting",
+#        "help": " -- examples: PEa - enable a (buzzer) [a=buzzer, b=overload bypass, j=power saving, K=LCD go to default after 1min, u=overload restart, v=overtemp restart, x=backlight, y=alarm on primary source interrupt, z=fault code record]",
+#        "response_type": ResultType.ACK,
+#        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+#        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+#        "regex": "PE(.+)$",
+#    },
+#   "PD": {
+#       "name": "PD",
+#       "description": "Set the disabled state of an Inverter setting",
+#       "help": " -- examples: PDa - disable a (buzzer) [a=buzzer, b=overload bypass, j=power saving, K=LCD go to default after 1min, u=overload restart, v=overtemp restart, x=backlight, y=alarm on primary source interrupt, z=fault code record]",
+#       "response_type": ResultType.ACK,
+#       "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+#       "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+#       "regex": "PD(.+)$",
+#   },
+#    "PF": {
+#        "name": "PF",
+#        "description": "Set Control Parameters to Default Values",
+#        "help": " -- example PF (reset control parameters to defaults)",
+#        "response_type": ResultType.ACK,
+#        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+#        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+#    },
+    "PGR": {
+        "name": "PGR",
+        "description": "Set Grid Working Range",
+        "help": " -- examples: PCR00 (set device working range to appliance), PCR01 (set device working range to UPS)",
+        "response_type": ResultType.ACK,
+        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+        "regex": "PGR(0[01])$",
+    },
+    "POP": {
+        "name": "POP",
+        "description": "Set Device Output Source Priority",
+        "help": " -- examples: POP00 (set utility first), POP01 (set solar first), POP02 (set SBU priority)",
+        "response_type": ResultType.ACK,
+        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+        "regex": "POP(0[012])$",
+    },
+    "POPLG": {
+        "name": "POPLG",
+        "description": "Set Device Operation Logic",
+        "help": " -- examples: POPLG00 (set Auto mode), POPLG01 (set Online mode), POPLG02 (set ECO mode)",
+        "response_type": ResultType.ACK,
+        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+        "regex": "POPLG(0[012])$",
+    },
+    "POPM": {
+        "name": "POPM",
+        "description": "Set Device Output Mode (for 4000/5000)",
+        "help": " -- examples: POPM01 (set unit 0 to 1 - parallel output), POPM10 (set unit 1 to 0 - single machine output), POPM02 (set unit 0 to 2 - phase 1 of 3), POPM13 (set unit 1 to 3 - phase 2 of 3), POPM24 (set unit 2 to 4 - phase 3 of 3)",
+        "response_type": ResultType.ACK,
+        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+        "regex": "POPM(\\d[01234])$",
+    },
+    "PPCP": {
+        "name": "PPCP",
+        "description": "Set Parallel Device Charger Priority (for 4000/5000)",
+        "help": " -- examples: PPCP000 (set unit 1 to 00 - utility first), PPCP101 (set unit 1 to 01 - solar first), PPCP202 (set unit 2 to 02 - solar and utility), PPCP003 (set unit 0 to 03 - solar only charging)",
+        "response_type": ResultType.ACK,
+        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
+        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+        "regex": "PPCP(\\d0[0123])$",
+    },
+    "PPVOKC": {
+        "name": "PPVOKC",
+        "description": "Set PV OK Condition",
+        "help": " -- examples: PPVOKC0 (as long as one unit has connected PV, parallel system will consider PV OK), PPVOKC1 (only if all inverters have connected PV, parallel system will consider PV OK)",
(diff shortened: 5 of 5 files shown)


docs/docker.md
```diff
@@ -53,3 +53,15 @@ services:
...
      - /dev/ttyUSB0:/dev/ttyUSB0
 ```
+
+Build Dev Container
+change to Git Directory 
+```
+docker build -t YOUR_TAG -f docker/dev/Dockerfile .
+```
+
+Stop mpp-sor container and Run Powermon one time
```
Saentist, line 63: Stop mpp-sor container >> Stop mpp-solar container

powermon/commands/response_definition.py
```diff
@@ -8,6 +8,7 @@
 class ResponseType(LowercaseStrEnum):
     ACK = auto()
     INT = auto()
+    TEN_INT = auto()
```
jblance, line 11: I dont want to head this way - I did start off like this for mppsolar logic, but soon ended up with many of this types
eg a quick look finds the below 
```
BigHex2Short:(r-30000)/10
BigHex2Short:r/10
BigHex2Short:r/1000
Hex2Int:r-40
BigHex2Short:r/100
BigHex2Short:(r-2731)/10
BigHex2Short:(r&0x7FFF)/100*(((r&0x8000)>>15)*2-1)
LittleHex2UInt:r/1000
LittleHex2Short:r/1000
```

and the list doesnt include the TEN_INT
Im thinking more of a template type, probably jinja2 where any base type can also have a template applied

powermon/commands/response_definition.py
```diff
@@ -229,6 +239,39 @@ def translate_raw_response(self, raw_value) -> str:
...
+                        data_value=str(value),
+                        data_unit=self.unit,
+                        device_class=self.device_class,
+                        state_class=self.state_class,
+                        icon=self.icon)]
+    
+    def get_description(self) -> str:
+        return self.description
+      
+# Same INT but base 10
```
rossandrews, line 254: Aren't all INTs base 10 by default? This is just dividing by 10

powermon/commands/response_definition.py
```diff
@@ -229,6 +239,39 @@ def translate_raw_response(self, raw_value) -> str:
...
+        self.index = index
+        self.description = description
+        self.unit = unit
+        self.device_class = device_class
+        self.state_class = state_class
+        self.icon = icon
+    
+    def is_valid_response(self, value) -> bool:
+        #Do we need to check if it's negative? Should we have max and min bounds?
+        return isinstance(value, int) / 10
```
rossandrews, line 269: you are dividing a bool by 10 here and I'm not sure what that will actually return. Currently this method is not used though and will likely be removed from the other response definitions.

powermon/commands/response_definition.py
```diff
@@ -229,6 +239,39 @@ def translate_raw_response(self, raw_value) -> str:
...
+    
+    def translate_raw_response(self, raw_value) -> str:
```
rossandrews, line 271: We should add a unit test for this logic. Currently I don't have any unit tests for the response definitions, I'll work on adding one so you can see an example.

powermon/commands/response_definition.py
jblance: But this will end up with a lot of extra ResponseTypes
I agree the different value types make sense, but when the answer needs math / transformation having a new ResponseType seems wrong.
I still prefer to have the ability to transform any value. Unit testing is still available with this approach
eg
Response is INT, unit test is straight forward
Response is INT with template/transformation - unit test is still straight forward (just need to supply raw response and template)
Else we end up with ResponseTypes like `BigHex2Short_Minus2731_Divide10` for perhaps a single result

rossandrews: How many types will we need to cover the different transformations? If it's less than 20 it's no big deal. And if the number starts to get unmanageable we can change course.

Another advantage of having different types is that if we need to change a type we only need to update one method instead of tracking down all the spots the logic appears.

rossandrews: We can also name the response types more descriptively and less literally. What does BigHex2Short_Minus2731_Divide10 do? convert Kelvin to Celsius? Then call it INT_KELVIN_TO_CELSIUS

jblance: > How many types will we need to cover the different transformations? If it's less than 20 it's no big deal. And if the number starts to get unmanageable we can change course.

Probably about 20, but I dont know tbh
> 
> Another advantage of having different types is that if we need to change a type we only need to update one method instead of tracking down all the spots the logic appears.

But your approach is the opposite, for this example if INT processing needs updating youll need to change both INT and TEN_INT (and 100_INT and 1000_INT ....)

The type is still INT, just to get the display value for some situations we need to divide by 10
Or more realistically the BigHex2Short has logic to convert to a value, but many of the values needs manipulation to give the 'correct' display - example above is to subtract 2731 before dividing by 10, why would you create another type for this single situation?

jblance: > We can also name the response types more descriptively and less literally. What does BigHex2Short_Minus2731_Divide10 do? convert Kelvin to Celsius? Then call it INT_KELVIN_TO_CELSIUS

First BigHex2Short decode sa 2 byte hexString to int (big endian coded) with `unpack(">h", hexString)[0]` 
And yes, the resulting value is 0.1K in units
There are also 'regular' BigHex2Short responses, and BigHex2Short responses that are 10mV units and some that are mV (or mA) units....

jblance: I still see a ResponseType and a transformation
INT and no transform
INT and divide by ten
BigHex2Short and no transform
BigHex2Short and divide by 10
BigHex2Short and divide by 10 and convert K to C
BigHex2Short and divide by 100, or 1000 etc

I just dont think the 2 should be combined into 1 type

rossandrews: We could do a more complex type structure instead of keeping it flat.

- Response
  - INT Response
    - INT divide 10 Response
  - BigHex2Short
    - BigHex2Short divide by 10
    - BigHex2Short divide by 10 and convert K to C

rossandrews: I will work on a pull request to demonstrate tonight (it's 3:50pm my time)

jblance: why not a single 'divide by 10' transform (which then can be used on INT, FLOAT, BigHex2Short, Hex2Int, LittleHex2Short, BigHex2Float, LittleHex2Float, LittleHex2UInt, LittleHex2Int....)

rossandrews: How about we introduce the concept of a unit of measurement. We can define which unit the device is giving us and what unit we want to output. Then we can store how to convert between units at the code level.

So BigHex2Short_Minus2731_Divide10 could be represented as:
 unit_from_device: DECIKELVIN, unit_in_response: CELSIUS

jblance: we'd need configuration items to specify desired units, and would need to have 10mV and 10mA etc as well as converters 10mV to V, 10mA to A etc. still seems like creating a lot of objects
What about response is 20231019 and we want to get year is 2023 month is 10 etc...
response is Wh and want KWh but only for one of the responses but not the others...

rossandrews: How about we pass in a method as a parameter? I think that's possible in python.

Something like UnitHelper.decikelvin_to_celsius Then we can test that logic easily. The type into or bytes will be passed in separately to inform how to decode.

jblance: getting more complex
ResponseType should be 'how is the data encoded, how do i get a value from it'
The 'this value means deci kelvin or mV, 10mV, 0.1V or mA' is separate - ideally we can convert it to 'standard' units after initially implementing this with a new ResponseType like the original PR I changed it to a template because of the increasing amount of types that were needed. Adding unit specific logic (A and V are different) would increase this even further
Either we go with PR like approach, or a transformation approach
PR-like would be
INT
deka_INT (need to /10 to get base)
deci_INT (need to *10 to get base)
note would also need deka_FLOAT, deka_BigHex2Short etc

temperature maybe should be dealt with separately as some would want C and some F (and maybe some K), so maybe that needs a config item and more complex types

transform would be to pass a transformation 'something' (string etc) that the value is transformed with (maybe a `transform=deka` in extra info)

remember we get responses like
`b"\xa5\x01\x90\x08\x02\x10\x00\x00uo\x03\xbc\xf3"`, with the decode looking like
```
           ["BigHex2Short:r/10", 2, "Battery Bank Voltage", "V"],
            ["BigHex2Short:r/10", 2, "acquistion", "V"],
            ["BigHex2Short:(r-30000)/10", 2, "Current", "A"],
            ["BigHex2Short:r/10", 2, "SOC", "%"],
```
and also like (pi18)
`b"^D0882300,217,2300,500,217,5000,5000,480,500,540,450,552,545,2,10,060,1,1,1,9,1,0,0,0,1,00\r",` 
```
           ["int", "AC output rating active power", "W", {"icon": "mdi:power-plug", "device-class": "power"}],
            ["10int", "Battery rated voltage", "V", {"icon": "mdi:battery-outline", "device-class": "voltage"}],
            ["10int", "Battery re-charge voltage", "V", {"icon": "mdi:battery-outline", "device-class": "voltage"}],
            ["10int", "Battery re-discharge voltage", "V", {"icon": "mdi:battery-outline", "device-class": "voltage"}],
            ["10int", "Battery under voltage", "V", {"icon": "mdi:battery-outline", "device-class": "voltage"}],
```
note this still has the 10int type in use (wonder if it still works...)

rossandrews: After sleeping on it I have a couple of thoughts:
1. We should replace the lists with dictionaries so we don't have to rely on the position in the list, should make future changes a bit easier and more readable for people new to the project

        "response": [
            [0, "Time until the end of absorb charging", ResponseType.INT, "sec"],
            [1, "Time until the end of float charging", ResponseType.INT, "sec"]
        ]

becomes

        "response": [
            {"index":0, "description":"Time until the end of absorb charging", "type":ResponseType.INT, "unit":"sec"},
            {"index":1, "description":"Time until the end of float charging", "type":ResponseType.INT, "unit":"sec"}
        ]

2. You are right that the Response data type and the transformation should be separate. We have a few different options:
**2.1.** We could split ResponseType into ResponseDecode and ResponseTransform. Our hypothetical BigHex2Short_Minus2731_Divide10 becomes ResponseDecode.BigHex2Short, ResponseTransform.DECIKELVIN_TO_CELSIUS
For millivolt to volt ResponseTranform could be MILLI, but I think MILLIVOLT_TO_VOLT is way more readable and worth the extra types. I prefer verbose easy to read code over terse code.
**2.2.** If you think there are too many transform types then we could have ResponseDecode, ResponseDeviceUnit, ResponseUnit. Instead of ResponseTransform.MILLIVOLT_TO_VOLT we use ResponseDeviceUnit.MILLIVOLT and ResponseUnit.VOLT. Then the code works out the translation and if it's valid. It is easy to unit test and is still readable without a ton of different types. We would prevent powermon from starting up if an invalid unit combo is presented.

jblance: I think the dict change is good - also allows the 'extra info' to be removed and added as additional keys
`{"index": 0, ..... "icon": "mdi:question", ...`
I still see ResponseDecode and ResponseType (what is the type of the data in the response) tbh
Probably can get rid of "type" as keyword too (seeing it is a python reserved word)
I guess we have the device response and the response we want for the results as separate concepts
if device response is ResponseXxxx then we should use something else for the other concept
going back to earlier thoughts, if we have ResponseType that allows us to understand the data, we can add a 'what units is the data returned in' and a 'what is the desired (or default) units for the result

ResponseType
ResponseUnits
DesiredUnits ??? (or something much better)

We then do transforms ResponseUnits to DesiredUnits but can also have overides (globally is easy, individually is harder but might be useful)

ie data is \x08, ResponseType is BigHex2Short (or whatever better name works), ResponseUnits are deka_volts, DesiredUnits are Volts, so process with deka_Volts to Volts 
Though this last one is a big if/switch statement

jblance: [sorry for all the 'stream of thoughts']

I also wonder if we can simplify ACK/SETTERS
response and test_responses are always the same, so can we just remove them and build into the logic?
```
"F": {
        "name": "F",
        "description": "Set Device Output Frequency",
        "help": " -- examples: F50 (set output frequency to 50Hz) or F60 (set output frequency to 60Hz)",
        "response_type": ResultType.ACK,
        "response": [[0, "Command execution", ResponseType.ACK, {"NAK": "Failed", "ACK": "Successful"}]],
        "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
        "regex": "F([56]0)$",
    },
```

rossandrews: So you are leaning toward 2.2? 
For serial number then we make ResponseUnit.SERIAL_NUMBER and DesiredUnit.STRING (or whatever) then put that logic in the transform between serial_number and string. If it differs between protocols then it could be ResponseUnit.PI18_SERIAL_NUMBER

jblance: yes leaning to 2.2
question is best way to determine what transform to apply (including the ability to override it)



> If it differs between protocols then it could be ResponseUnit.PI18_SERIAL_NUMBER

yes this works - custom or specific ResponseUnits can deal with most/all weird situations (as long as we have a good solution to the Q above)

rossandrews: I would write objects for each DesiredUnit (OutputUnit? TranslatedUnit? DecodedUnit?) that knows how to convert from the ResponseUnit. There will only be a few types of DesiredUnits and they can hold all that logic.

Instead of needing a way to override, adding a new ResponseUnit and adding corresponding code to the DesiredUnit should work nicely. I really don't think any code should go into the protocols, keeping the code in proper classes will make maintenance easier.

jblance:
Away from computer so excuse typos

Override is more so an individual could choose to use Fahrenheit instead of
Celsius, the transform is easy enough but we need some way (in the config
file) to tell the code the DesiredUnit is (in this example) Fahrenheit not
Celsius

powermon/commands/response_definition.py
rossandrews: > Away from computer so excuse typos
>
>Override is more so an individual could choose to use Fahrenheit instead of
>Celsius, the transform is easy enough but we need some way (in the config
>file) to tell the code the DesiredUnit is (in this example) Fahrenheit not
>Celsius

@jblance you responded with a higher level comment but I'll respond here. We can do that at the display level, the DesiredUnit could be Temperature that outputs C, F or K depending on config.

jblance:
had a first cut at the pi18 with the much changed powermon protocol format
i noticed that most of the commands in the above PR for the pi18 were straight out of pi30, so have inherited and just added PIRI, not sure if this is correct or not
scratch that, pi18 commands are different to pi30, so only the 1 PIRI command in the PR is valid anyway

PR #414, opened 2023-10-20
Update json_mqtt.py

sante85:
Add tag information to topic


rossandrews:
Good catch and bugfix, could you add a unit test? There are several format unit tests you can use as examples, like test_format_htmltable.py

sante85:
for me is difficult now. You have a suggests for speedup this phase?

Il giorno ven 20 ott 2023 alle ore 18:50 Ross Andrews <
***@***.***> ha scritto:

sante85:
When do you think to make a release of python Pi repository?

Il dom 22 ott 2023, 03:58 John Blance ***@***.***> ha scritto:

jblance:
Should be there now

PR #416, opened 2023-10-23
Adding default None for config that is optional

rossandrews:
Also adding a unit test that confirms the config files pass the config_model validation


jblance:
It now seems that any field that allows none but is not present causes validation to fail - is this due to newer pydantic?
Do we need default = None for all optional fields or is there another way?

rossandrews:
I am not sure but it wasn't happening when you first added it. The default field seems to work well though and we can be more explicit.

PR #418, opened 2023-10-28
Adding units to protcols

rossandrews:
@jblance implementing some of the points we discussed on the changes to the protocol. For terminology I was thinking a Response is what comes from the device. The Result is what you get back from executing a command and it has one or many Readings that each have a type. I also introduced the idea of a Parameter that is separate from a reading and contains information that is not provided by executing the command.

Most of the changes are naming related, I put together tests/powermon_tests/unit/test_abstract_protocol.py to demo the changes. Currently this code doesn't compile since I haven't applied the changes to the protocols. Figured we should come to a common understanding before refactoring those.


jblance:
Wow, a lot of work here. Awesome
Might be worth updating a single command in pi30 to demo your intent

jblance:
> Response is what comes from the device. The Result is what you get back from executing a command and it has one or many Readings that each have a type

So Response is the raw response
Result is the processed 'response' and it has Readings (each one is a single thing, like Battery Voltage)
Command is the object with the details about what is being sent (inc command.code which is the specific command eg QPIGS)
Is that what you mean?

mppsolar/helpers.py
jblance, line 1: We should move the powermon stuff to powermon and leave mppsolar stuff isolated

powermon/device.py
```diff
@@ -58,7 +58,7 @@ def add_command(self, command: Command) -> None:
         if command is None:
             return
         # get command definition from protocol
-        command.set_command_definition(self.port.protocol.get_command_definition(command.code))
+        command.set_command_definition(self.port.protocol.get_command_with_command_string(command.code))
```
jblance, line 61: `get_command_from_string` or `definition_from_string` (to align more with from_config, to_dto etc)

tests/powermon_tests/unit/test_abstract_protocol.py
```diff
@@ -0,0 +1,108 @@
...
+                    {"name":"date", "description":"Date for query", "parameter_type":ParameterType.DATE, "parameter_format":ParameterFormat.YYYYMMDD}
+                ],
+                "test_responses": [
+                    b"(00238800!J\r",
+                ],
+                "regex": "QED(\\d\\d\\d\\d\\d\\d\\d\\d)$",
+            }
+        }
+        
+        protocol.add_command_definitions(qed_command_definition_new, "QUERY")
```
jblance, line 90: Do we still need QUERY / SETTER separation, should these be better defined, more separated, removed entirely...
Always had (in the back of my mind) a more 'understandable' or interchangable command syntax (--getStatus was initial attempt), eg `getBatteryVoltage` etc - still dont really see a way to do that sensibly tho

tests/powermon_tests/unit/test_abstract_protocol.py
```diff
@@ -0,0 +1,108 @@
...
+                ],
+                "regex": "QED(\\d\\d\\d\\d\\d\\d\\d\\d)$",
+            }
+        }
+        qed_command_definition_new = {
+            "QED": {
+                "name": "QED",
+                "description": "Daily PV Generated Energy Inquiry",
+                "help": " -- display daily generated energy, format is QEDyyyymmdd",
+                "result_type": ResultType.INDEXED,
```
jblance, line 76: Should this be more targetted? ResultType.SINGLE or similar

tests/powermon_tests/unit/test_abstract_protocol.py
```diff
@@ -0,0 +1,108 @@
...
+                "reading_definitions": [
+                    {"index":0, "description":"PV Generated Energy for Day", "reading_type":ReadingType.WATT_HOURS, "response_type":ResponseType.INT, "icon": "mdi:solar-power", "device-class": "energy", "state_class": "total"},
+                ],
+                "parameters": [
+                    {"name":"date", "description":"Date for query", "parameter_type":ParameterType.DATE, "parameter_format":ParameterFormat.YYYYMMDD}
```
jblance, line 81: But how do we use this
The info readings were to 'add' readings that gave extra (perhaps unneeded) info
This parameter exposes this info to the code, but not to the Result (and therefore not to the output streams)

tests/powermon_tests/unit/test_abstract_protocol.py
```diff
@@ -0,0 +1,108 @@
...
+                "regex": "PBT(0[012])$",
+            },
+        }
+        pbt_command_definition_new = {
+            "PBT": {
+                "name": "PBT",
+                "description": "Set Battery Type",
+                "help": " -- examples: PBT00 (set battery as AGM), PBT01 (set battery as FLOODED), PBT02 (set battery as USER)",
+                "result_type": ResultType.ACK,
+                "reading_definitions": [{"index":0, "decription":"Command execution", "reading_type":ReadingType.ACK, "response_type":ResponseType.ACK}],
```
jblance, line 31: I think this and the line below are redundant and can be determined by ResultType.ACK

tests/powermon_tests/unit/test_abstract_protocol.py
```diff
@@ -0,0 +1,108 @@
...
+                "test_responses": [b"(NAK\x73\x73\r", b"(ACK\x39\x20\r",],
+                "regex": "PBT(0[012])$",
+            }
+        }
+        protocol.add_command_definitions(pbt_command_definition_new, "QUERY")
```
jblance, line 36: surely "QUERY" here isnt correct, "SETTER" or something....

mppsolar/helpers.py
rossandrews: I guess this is referenced in the powermon code, this was changed as part of a rename. I'll undo it.

tests/powermon_tests/unit/test_abstract_protocol.py
rossandrews: I've been thinking about this as well. I think we can define a device type, like solar controller, then we can define a command and reading that will fulfill the interface. So for each protocol we implement a getBatteryVoltage and then say which command and which reading will return that result.

tests/powermon_tests/unit/test_abstract_protocol.py
rossandrews: Sure

tests/powermon_tests/unit/test_abstract_protocol.py
rossandrews: Currently I am not using the query or setter. If we need it it could be included in the command configuration.

rossandrews:
> > Response is what comes from the device. The Result is what you get back from executing a command and it has one or many Readings that each have a type
> 
> So Response is the raw response Result is the processed 'response' and it has Readings (each one is a single thing, like Battery Voltage) er, where does Parameter fit? Command is the object with the details about what is being sent (inc command.code which is the specific command eg QPIGS) Is that what you mean?

Yes this is the overall structure. Parameters are in the command and can be included in the Result as a separate bit of information to help. I'll modify the Result to demonstrate.

tests/powermon_tests/unit/test_abstract_protocol.py
jblance: do we need it? 
i guess SETTER = ACK response only....

tests/powermon_tests/unit/test_abstract_protocol.py
jblance: other devices (BMS, Shunt ...) also have batteryVoltage
Maybe more a 'supported commands' for each protocol approach

powermon/commands/result.py
```diff
@@ -43,12 +43,12 @@ def set_device_id(self, device_id):
...
-    def get_responses(self) -> list[Response]:
-        return self.responses
-
-    def add_responses(self, responses: list[Response]) -> bool:
-        self.responses.extend(responses)
+    
+    def get_responses(self) -> list[Reading]:
+        return self.readings
+    
+    def add_responses(self, responses: list[Reading]) -> bool:
```
jblance, line 50: i guess these 'responses' should be called 'readings' too?

rossandrews: yes, I've missed it in a few spots. Working on cleaning it up.

powermon/protocols/abstractprotocol.py
```diff
@@ -123,65 +123,16 @@ def get_command_with_command_string(self, command) -> CommandDefinition:
...
+    def check_response_and_trim(self, response: str) -> str:
         """
         Simplest validity check, CRC checks should be added to individual protocols
         """
-        if result.raw_response is None:
-            result.is_valid = False
-            result.error = True
-            result.error_messages.append("failed validity check: response was empty")
+        if response is None:
+            raise ValueError("Response is None")
```
jblance, line 133: I wonder if we should expand the exceptions we use - powermon.Errors.InvalidResponse etc, so we can deal with them 'more properly'

tests/powermon_tests/unit/test_abstract_protocol.py
jblance: also shoud QUERY be GETTER to be more in line with get_ set_ python way?

powermon/protocols/abstractprotocol.py
rossandrews: Yeah we can include more errors when we get to the point of handling expected errors gracefully.

tests/powermon_tests/unit/test_abstract_protocol.py
rossandrews: I think we should move it into the command definition. Maybe just include it in the ResultType. ResultType.COMMAND which will have a reading of an ACK. Then the other types will be getters.

jblance: yeah not sure we really need the separation / extra 'definition' item
anything we can remove from the definition the better imo

powermon/commands/result.py
```diff
@@ -1,40 +1,48 @@
...
+from powermon.commands.parameter import Parameter
+from powermon.commands.reading_definition import ReadingDefinition
 from powermon.dto.resultDTO import ResultDTO
 
 log = logging.getLogger("result")
 
 class ResultType(Enum):
     DEFAULT = auto()
     ACK = auto()
+    COMMAND = auto()
```
jblance, line 14: not another command! :-P
a variation on ACK might be less confusing?

jblance: think we need to define the options
1, ack type (success / fail of command)  ACK, COMMAND
2, a series of values, the place (1st, 2nd etc) in relation to overall defines what the value relates to, INDEXED, DEFAULT
3, a long 'string', where the position (eg characters 3-5) defines that the value relates to POSITIONAL??
4, a series of values that are all related to a single answer MULTIVALUED
5, key: value data (multiple per 'response') where the key defines what the value relates to  ??
6, multiframe (cant remember how this works off hand) ??
7, single value, SINGLE

jblance: DEFAULT should go now, was a temp while the protocols were migrated to new approach in mppsolar code

rossandrews: Yeah, we have enough commands, I'll rename it.

I will add more ResultTypes when I start converting the protocols. Do you like the pattern enough that I should start applying it to the whole protocol? It will take a little while.

jblance: yeah, i think the pattern is great - better to have more types than lots of complexity in decode
I'm keen to merge as soon as possible (even unfinished) so I can make some changes without running into merge problems - are ok to merge soon?
Also have you seen the port PR (just some tidy atm, but Im wanting to play with BLE)??

rossandrews: If you are okay with it being broken for a bit you can merge anytime. We can work together on fixing it up.

PR #419, opened 2023-11-03
Powermon ports and other tweaks

jblance:


docker/powermon.yaml
```diff
@@ -16,7 +16,7 @@ commands:
     outputs:
       - type: mqtt
         format: 
-        - type: mqttmessage
+          type: mqttmessage
```
jblance, line 19: this change is to match current config_model, question is should we allow multiple formats (in which case we'd want a list with a `-`) and need to change some logic

powermon/ports/__init__.py
```diff
@@ -1,42 +1,66 @@
...
-    if portType is None:
+    if port_type is None:
         return None
 
     # build port object
-    if portType == PortType.SERIAL:
-        portObject = SerialPort.fromConfig(config=port_config)
-    elif portType == PortType.USB:
-        portObject = USBPort.fromConfig(config=port_config)
+    match port_type:
```
jblance, line 49: I'm adopting the match pattern in preference to if, elif, else blocks where possible

rossandrews (approved):
Looks good, I am looking forward to trying out some bluetooth. I bought a Daly BMS with an active balancer I want to build a 24V lifepo4 battery with.

jblance:
> Looks good, I am looking forward to trying out some bluetooth. I bought a Daly BMS with an active balancer I want to build a 24V lifepo4 battery with.

yeah I have a DALY as well (48V lifepo4 bank), ble needs a lot of work as the original was very specific to JKBMS

PR #420, opened 2023-11-03
Protocol wip

jblance:
just looking to get QID working in PI30 with new approach
Doing PR to allow discussion with @rossandrews


powermon/protocols/pi30.py
```diff
@@ -245,6 +246,16 @@
 }
 
 QUERY_COMMANDS = {
+    "QID": {
+        "name": "QID",
+        "description": "Device Serial Number inquiry",
+        "help": " -- queries the device serial number",
+        "result_type": ResultType.INDEXED,
+        "reading_definitions": [{
+            "index": 0, "description": "Serial Number", "reading_type": ReadingType.ACK, "response_type": ResponseType.BYTES
```
jblance, line 255: ACK is just temp

rossandrews:
Sorry missed the initial pull request, had a busy weekend. I was working on the Q1 command today, I'll add some comments tomorrow and put in my PR as well.

jblance:
> Sorry missed the initial pull request, had a busy weekend. I was working on the Q1 command today, I'll add some comments tomorrow and put in my PR as well.

No problem, I just happened to have some time to look at the work that you've done.
One thing to watch is that I want the code to work across different devices and protocols. This means the protocols need to do a lot of the work, eg even splitting the response into results differs by protocol

rossandrews:
I've been thinking, it's very time consuming changing the structure of these protocols and we are only doing two of them. To make any future changes easier putting all the protocol configuration into a SQLite DB would really help. If you are not familiar with SQLite it's built for situations like this and is used widely (https://www.sqlite.org/index.html). It would also decouple protocol configuration from the code. In the future powermon could get the latest protocol configuration without needing a code update.

A human readable version of the protocol can be made available through a command or through the API. Currently the API does get the protocol information I just haven't spent any time making it look nice.

rossandrews:
This code looks great, I will check it out and merge my changes into it. Merge it in whenever you are ready.

jblance:
> I've been thinking, it's very time consuming changing the structure of these protocols and we are only doing two of them. To make any future changes easier putting all the protocol configuration into a SQLite DB would really help. If you are not familiar with SQLite it's built for situations like this and is used widely (https://www.sqlite.org/index.html). It would also decouple protocol configuration from the code. In the future powermon could get the latest protocol configuration without needing a code update.
> 
> A human readable version of the protocol can be made available through a command or through the API. Currently the API does get the protocol information I just haven't spent any time making it look nice.

Yeah the protocol config is pretty clunky. I feel there is still a need for some python functions per protocol but the COMMAND dict / list etc is annoying
Ive been trying to keep things simple (er well as simple as possible) and am a little alergic to DBs in general but now is the right time to come up with a better protocol approach. If you can suggest a DB schema that makes sense (for all protocols, including the BLE stuff) I happy to go that way (especially if it reduces the amount of work to update the protocols - the last thing I was looking at was a way to remove some of the redundant config for ResultType.ACK

PR #422, opened 2023-11-08
Updating command q1

rossandrews:
Working on getting the ReadingType, ResponseType pattern in a good state before trying out any DB work. Running commands via USB not working at the moment, still a WIP.


powermon/commands/command.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, code: str, commandtype: str, outputs: list[AbstractOutput], t
         log.debug(self)
 
     def build_result(self, raw_response=None, protocol=None) -> Result:
-        log.debug("build_result: code:%s, command_definition:%s", self.code, self.command_definition)
+        log.debug(f"build_result: code:{self.code}, command_definition:{self.command_definition}")
```
jblance, line 43: Ive been trying to move f-strings to lazy expansion in debug messages (as per pylint recommendations)

powermon/protocols/pi30.py
```diff
@@ -826,6 +786,48 @@
...
+        "description": "Device Serial Number inquiry",
+        "help": " -- queries the device serial number",
+        "result_type": ResultType.SINGLE,
+        "reading_definitions": [{"description": "Serial Number", "reading_type": ReadingType.MESSAGE, "response_type": ResponseType.STRING}],
+        "test_responses": [b"(9293333010501\xbb\x07\r"],
+    },
+    "Q1": {
+        "name": "Q1",
+        "description": "Q1 query",
+        "result_type": ResultType.INDEXED,
```
jblance, line 801: Not sure this is actually indexed, it really is position in result (which was positional in mppsolar before I started to migrate protocols decode.
We should use 'correct' approach from the start since we're changing everything

powermon/protocols/pi30.py
```diff
@@ -826,6 +786,48 @@
...
+            {"index":3, "reading_type":ReadingType.MESSAGE, "response_type":ResponseType.OPTION, "description":"AllowSccOnFlag", "options":{"00":"SCC not allowed to charge", "01":"SCC allowed to charge"}},
+            {"index":4, "reading_type":ReadingType.AMPERAGE, "response_type":ResponseType.INT, "description":"ChargeAverageCurrent"},
+            {"index":5, "reading_type":ReadingType.TEMPERATURE, "response_type":ResponseType.INT, "description":"SCC PWM temperature", "device-class": "temperature"},
+            {"index":6, "reading_type":ReadingType.TEMPERATURE, "response_type":ResponseType.INT, "description":"Inverter temperature", "device-class": "temperature"},
+            {"index":7, "reading_type":ReadingType.TEMPERATURE, "response_type":ResponseType.INT, "description":"Battery temperature", "device-class": "temperature"},
+            {"index":8, "reading_type":ReadingType.TEMPERATURE, "response_type":ResponseType.INT, "description":"Transformer temperature", "device-class": "temperature"},
+            {"index":9, "reading_type":ReadingType.TEMPERATURE, "response_type":ResponseType.INT, "description":"GPIO13"},
```
jblance, line 812: pretty sure this is not a temperature

powermon/protocols/pi30.py
```diff
@@ -826,6 +786,48 @@
...
+        "result_type": ResultType.SINGLE,
+        "reading_definitions": [{"description": "Serial Number", "reading_type": ReadingType.MESSAGE, "response_type": ResponseType.STRING}],
+        "test_responses": [b"(9293333010501\xbb\x07\r"],
+    },
+    "Q1": {
+        "name": "Q1",
+        "description": "Q1 query",
+        "result_type": ResultType.INDEXED,
+        "reading_definitions": [
+            {"index":0, "reading_type":ReadingType.TIME, "response_type":ResponseType.INT, "description":"Time until the end of absorb charging"},
```
jblance, line 803: I was going around in circles on response_type and reading_type - this looks like the correct 'landing', ie response_type is limited to what the data is
response_type will need to include the struct.unpack types when we get there (BigHextoInt etc) but Im liking where youve got to

powermon/protocols/ved.py
```diff
@@ -132,7 +132,7 @@ def __str__(self):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__()
         self._protocol_id = b"VED"
-        super().add_command_definitions(COMMANDS, "QUERY")
+        super().add_command_definitions(COMMANDS)
```
jblance, line 135: self.add_commands.... ?

powermon/commands/reading_definition.py
```diff
@@ -326,110 +375,12 @@ def reading_from_raw_response(self, raw_value) -> list[Reading]:
...
-    def __init__(self, index: int, description: str, unit: str, device_class: str = None, state_class: str = None, icon: str = None):
-        self.index = index
-        self.description = description
-        self.unit = unit
-        self.device_class = device_class
-        self.state_class = state_class
-        self.icon = icon
+class ReadingDefinitionTemperature(ReadingDefinition):
+    def __init__(self, index: int, name: str, description: str,  response_type: ResponseType, device_class: str = None, state_class: str = None, icon: str = None):
+        #TODO: find a way to make the unit configurable
```
jblance, line 380: Also need to consider how to 'normalize' the input - ie remember the response that was in kelvin (and needing /10 to get standard K)

powermon/commands/command.py
rossandrews: It wasn't printing properly during an exception being thrown so it wasn't very helpful. I can switch it back, should work if there isn't an exception.

powermon/protocols/pi30.py
rossandrews: I'm glad you like it, I think it's turned out to be pretty readable. Only part that gets me sometimes is that Result, Reading, and Response are all too similar in spelling when I'm reading quickly.

powermon/commands/reading_definition.py
```diff
@@ -272,7 +309,9 @@ def __init__(
         state_class: str = None,
         icon: str = None
     ):
-        super().__init__(index, name, response_type, "Wh", description, device_class, state_class, icon)
+        super().__init__(index, name, response_type, description, device_class, state_class, icon, unit="Wh")
+        if response_type not in [ResponseType.INT]:
```
rossandrews, line 313: In the constructor of each Reading type it will check if the provided ResponseType is supported. What type of error should I raise when this happens?

jblance: i think we should use custom exceptions more - to make future handling easier

powermon/commands/reading_definition.py
rossandrews: My current thinking is that it will be ResponseType.INT_DECIKELVIN (maybe just DECIKELVIN) or something like that then the ReadingType.TEMPERATURE will handle the conversion logic.

jblance: Agree. I think the first option so we could have other types eg uint or hex or bytes or whatever if needed

powermon/ports/abstractport.py
```diff
@@ -58,7 +59,8 @@ def run_command(self, command: Command) -> Result:
 
         # open port if it is closed
         if not self.is_connected():
-            self.connect()
+            if self.connect() == -1:
```
jblance, line 62: self.connect returns bool

powermon/ports/usbport.py
```diff
@@ -51,6 +55,9 @@ def disconnect(self) -> None:
         self.port = None
 
     def send_and_receive(self, command: Command) -> Result:
+        if not self.is_connected():
+            log.warning("USBPort not connected")
```
jblance, line 59: or try reconnect?

powermon/commands/result.py
```diff
@@ -120,8 +120,10 @@ def split_responses(self, response) -> list:
 
     def validate_and_translate_raw_value(self, raw_value: str, index: int) -> list[Reading]:
         if len(self.reading_definitions) <= index:
-            raise IndexError(f"Index {index} is out of range for command {self.command_code}")
-        reading_definition: ReadingDefinition = self.reading_definitions[index]
+            log.debug(f"Index {index} is out of range for command {self.command_code}")
```
jblance, line 123: no f strings for log messages

powermon/device.py
```diff
@@ -103,7 +103,7 @@ def run(self, force=False):
                     result: Result = self.port.run_command(command)
                 except Exception as exception:  # pylint: disable=W0718
                     log.error("Error decoding result: %s", exception)
-                    result = Result(command_code=command.code, result_type=ResultType.ERROR, raw_response="")
+                    result = Result(command_code=command.code, result_type=ResultType.ERROR, raw_response=b"Error decoding result")
```
jblance, line 106: this is where we should get more specific on error handling (eventually), eg PortNotConnected vs UnableToDecode type separation

powermon/commands/reading_definition.py
jblance: maybe a match case construct (in the translate_from_raw function), with `case _:` covering the error state and each other state doing the translation 

InvalidResponse or similar error, so we can propagate the error up and deal without crashing

powermon/commands/reading_definition.py
```diff
@@ -121,96 +120,134 @@ def from_config(cls, reading_definition_config: dict, i) -> "ReadingDefinition":
...
+        description: str,
+        device_class: str = None,
+        state_class: str = None,
+        icon: str = None,
+        unit: str = ""
+    ):
+        super().__init__(index, name, response_type, description, device_class, state_class, icon, unit=unit)
+
+        
+    def is_valid_response(self, value) -> bool:
```
jblance, line 233: do we need this as a separate function? can it be part of `translate_raw` possibly passing ResponseType so the processing can do more work or raise exception for invalid

powermon/protocols/pi30.py
jblance: Im thinking this is `ResultType.POSITION` and we has `"position": 0` and index an be kept for results like ved `"response": [ ["V", "Main or channel 1 battery voltage", "V", "float:r/1000"],`

powermon/commands/result.py
```diff
@@ -78,7 +89,7 @@ def decode_response(self, raw_response) -> list[Reading]:
 
         # Process response based on result type
         match self.result_type:
-            case ResultType.SINGLE:
+            case ResultType.SINGLE | ResultType.ERROR:
```
jblance, line 92: probably needs own case for .ERROR

powermon/protocols/pi30.py
jblance: yes, and we could argue in circle on which is which for ever, eg I still see response as what the device sends (currently result though i dont know what the smaller response piece should be in this way of thinking (full response and response item but thats too many letters..)
Anyway current way works, just need to keep it consistent...

rossandrews:
@jblance Sorry about disappearing for a bit, I'll get back into it soon. I will be on a trip from December 1 - 16 so will have these latest changes all wrapped up by then.

powermon/outputs/abstractoutput.py
```diff
@@ -1,43 +1,51 @@
...
-    def get_topic(self) -> str:
-        return ""
-
     @abstractmethod
-    def process(self, result: Result):
-        pass
-
-    def set_mqtt_broker(self, mqtt_broker: MqttBroker):
-        pass
+    def process(self, result: Result, mqtt_broker=None):
```
rossandrews, line 27: changed this from device to mqtt_broker, easier to test and the same logic.

jblance: for hass we use the device details to populate the 'device' for the mqtt messages
i was debating passing device and mqtt_broker separately or adding mqtt_broker to device, not really happy with either, but need both somehow for certain outputs

rossandrews: hass is a format not an output, it already has device.

powermon/commands/result.py
rossandrews: I'll split it out but it will be the same for now.

powermon/commands/reading_definition.py
rossandrews: I was thinking it would contain validation checks, like temperature has to be a float and can't be negative if it's in Kelvin. Having it separate also helps with testing eventually. Currently it's under utilized and I'll remove them to make the code cleaner.

powermon/device.py
rossandrews: I agree we need work on the error handling, but lets save it for a future PR.

powermon/ports/usbport.py
rossandrews: Lets add this to the error handling work. Should we make a ticket with a list of error handling tasks?

powermon/device.py
jblance: agree

powermon/outputs/abstractoutput.py
jblance: ok great

rossandrews:
I am going to start on a SQLite version of the protocols tomorrow.

powermon/protocols/pi30.py
rossandrews: what should it be?

jblance:
> I am going to start on a SQLite version of the protocols tomorrow.

Im not sure on the DB approach, how do we keep the DB up to date, deal with different versions etc
Is the DB built from scratch for each update? Else how do changes get incorporated, how do deltas get sorted etc

rossandrews:
> > I am going to start on a SQLite version of the protocols tomorrow.
> 
> Im not sure on the DB approach, how do we keep the DB up to date, deal with different versions etc Is the DB built from scratch for each update? Else how do changes get incorporated, how do deltas get sorted etc

I will write up my plan before starting to implement it. The SQLite db will live with the code just like the protocols currently do. It will just be easier to query and easier to build relationships between concepts. Hopefully it will also make it easier to modify commands so other DIYers can try stuff out without getting into the code.

jblance:
So are you ready for this to be merged? (assume any DB work to be a separate PR)

rossandrews:
Sorry missed your last comment, I will work on the DB stuff in a new PR. If you are happy with the progress on this I'm ready to merge.

powermon/device.py
```diff
@@ -49,13 +50,19 @@ def __init__(self, name: str, device_id: str = "", model: str = "", manufacturer
...
         """add a command to the devices' list of commands"""
+        log.debug("Adding command: %s", command)
         if command is None:
             return
         # get command definition from protocol
-        command.set_command_definition(self.port.protocol.get_command_with_command_string(command.code))
+        command_definition = self.port.protocol.get_command_with_command_string(command.code)
+        if command_definition is None:
+            log.error("Cannot find command code: %s, in protocol: %s", command.code, self.port.protocol.get_protocol_id())
+            raise RuntimeError(f"Invalid command code: {command.code}, in protocol: {self.port.protocol.get_protocol_id()}")
```
jblance, line 64: i wonder whether its better to raise the error (perhaps even a custom one like InvalidCommandError) in the protocol.get_command_with_command_string function (closer to logic and simpler here), then we can just use `command.set_command_definition(self.port.protocol.get_command_with_command_string(command.code))`

powermon/formats/table.py
```diff
@@ -108,3 +110,20 @@ def format(self, result: Result) -> list[str]:
             _result.append("\u255a" + ("\u2550" * (width_p + 1)) + "\u2567" + ("\u2550" * (width_v + 1)) + "\u2567" + ("\u2550" * (width_u + 1)) + "\u255d")
         # _result.append("\n")
         return _result
+
+
+def get_max_response_length(responses: list[Reading]):
```
jblance, line 115: why is this here? should use the imported ones

powermon/outputs/abstractoutput.py
jblance: > hass is a format not an output, it already has device.

how?

powermon/ports/usbport.py
jblance: > Lets add this to the error handling work. Should we make a ticket with a list of error handling tasks?

makes sense, need to track stuff somewhere (or add # TODO: items

PR #430, opened 2023-12-11
fix: Usage of deprecated `logging.warn()`

BoboTiG:


jblance:
thanks for your work!

PR #431, opened 2023-12-11
fix: deprecation warning about an invalid escape sequence in `testio.py`

BoboTiG:


jblance:
thanks for your contribution

PR #433, opened 2023-12-12
fix: CLI argument parsers types

BoboTiG:
- The default type is `str`, no need to set it then;
- 2 cases where the type was set as `str` when it should be `int`.


jblance:
nice improvement

PR #435, opened 2023-12-12
feat: Introduce the `prom_push` output

BoboTiG:
It will push metrics to a running PushGateway, and Prometheus will need to be configured to pull metrcis from there.

The default URL can be changed via the `push_url` from the config file, example:

```ini
[Inverter_1]
port=/dev/ttyUSB0
protocol=PI30M045
command=QPGS0
outputs=prom_push
push_url=http://example.org/metrics/job/foo
```

Or via the `--pushurl` CLI argument.

On the Prometheus side, add a new scrape config:

```yaml
scrape_configs:
  # (...)

  - job_name: pushgateway
    honor_labels: true
    static_configs:
      - targets:
          - localhost:9091
```

And that's it!

Closes #434.

### DIFF
--- a/mppsolar/__init__.py
+++ b/mppsolar/__init__.py
@@ -153,6 +153,14 @@ def main():
         default="mppsolar",
     )
     parser.add_argument(
+        "--pushurl",
+        help=(
+            "Any server used to send data to (PushGateway for Prometheus, for instance), "
+            "(default: http://localhost:9091/metrics/job/pushgateway)"
+        ),
+        default="http://localhost:9091/metrics/job/pushgateway",
+    )
+    parser.add_argument(
         "-c",
         "--command",
         nargs="?",
@@ -261,6 +269,7 @@ def main():
     excl_filter = args.exclfilter
     keep_case = args.keepcase
     mqtt_topic = args.mqtttopic
+    push_url = args.pushurl
 
     _commands = []
     # Initialize Daemon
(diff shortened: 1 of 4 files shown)


BoboTiG:
Note: I refactored slightly the `prom` class to be reduce the maount of duplicated code.

pyproject.toml
```diff
@@ -29,6 +29,7 @@ fastapi-mqtt = ">=1.1"
 # jinja2 = {version = ">=3.0.0", optional = true}
 # fastapi-mqtt = {version = ">=1.0.7", optional = true}
 sqlalchemy = "^2.0.15"
+requests = ">=2.31.0"
```
BoboTiG, line 32: For now it should do the trick, if you are OK with that.
In a near future, it could be interesting to split dependencies based on outputs. For instance, when using the `prom` output, installing FastAPI or even uvicorn doens't make sense. As a user, I know what output(s) I'll use, so it could be a good way of installing only strictly required third-party modules. WDYT?

jblance: yeah i was think about this - i even started implemented, but didnt really get it done (especially some of the 'heavier' dependencies

mppsolar/__init__.py
```diff
@@ -152,6 +152,14 @@ def main():
         help="Mongo db name (default: mppsolar)",
         default="mppsolar",
     )
+    parser.add_argument(
```
jblance, line 155: the need for additional config parameters is one of the things Im trying to address in the powermon version (using a config yaml to allow different and flexibile config for each item)

pyproject.toml
BoboTiG: I could give a try if you are not yet working on it.

jblance: Im not working on it at the moment

PR #428, opened 2023-12-04
New PI30MAX protocol for 044 model (Axpert Max 7.2k)

gianfrdp:
I added QBMS command and adapted some other commands to Axpert Max 7.2 (model 044)

PR #440, opened 2023-12-14
feat: Use extras to install specific dependencies

BoboTiG:
The only missing piece is MQTT, it is way to intriqued everywhere.
So for now, the `paho-mqtt` dependency will be installed by default.

At some point, it will be cool to be able to get rid of it and install it via:

```shell
$ python -m pip install 'mppsolar[mqtt]'
```

Also, I added very primary tests to see if all protocols and outputs can be
imported even when a depedency is missing.

A proper message will be printed
in that case (except for `esp32io.py` and the `machine` import, I guss this is very specific?).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,12 @@ Currently has support for:
 
 ## Install ##
 - `pip install mppsolar` (minimal install), or
-- `pip install mppsolar[ble]` (install including BLE support aka jkbms), or
+- `pip install mppsolar[api]` (install server stuff), or
+- `pip install mppsolar[ble]` (for Bluetooth support aka jkbms), or
+- `pip install mppsolar[mongo]` (for MongoDB output), or
+- `pip install mppsolar[pgsql]` (for PostgreSQL output), or
+- `pip install mppsolar[push]` (for Pometheus PushGateway output), or
+- `pip install mppsolar[systemd]` (for `--daemon` on Linux), or
 - `docker pull jblance/mppsolar:latest` (docker install)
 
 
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -21,7 +21,12 @@ Currently has support for:
 
 ## Install ##
 - `pip install mppsolar` (minimal install), or
-- `pip install mppsolar[ble]` (install including BLE support aka jkbms), or
+- `pip install mppsolar[api]` (install server stuff), or
```
BoboTiG, line 24: I am not proud about "server stuff", so if you come up with something better, that would be appreciated :)

tests/mppsolar_tests/test_outputs.py
```diff
@@ -0,0 +1,23 @@
+import importlib
+import unittest
+from pathlib import Path
+
+HERE = Path( __file__).parent
+OUTPUTS_DIR = HERE.parent.parent / "mppsolar" / "outputs"
+OUTPUTS = sorted(
+    file.stem for file in OUTPUTS_DIR.glob("*.py")
+    if file.stem != "__init__"
+)
```
BoboTiG, line 10: We can't use `list_output()` here, for two reasons:
- it doesn't work from tests, it seems the fact that it's not packaged is messing the whole logic
- more importantly, we need to really try all outputs, without skipping ones that are not available

Same remark for the protocol test file below.

gianfrdp:
This PR breaks my previous PR #428

for file mppsolar/protocols/abstractprotocol.py

firefox_2023-12-15_15-17-37

BoboTiG:
Then a `# noqa: F401` would have been wise. I'll open a PR to ix that.

PR #446, opened 2023-12-19
Update pi30m045.py

cordelster:
FIX: Make "MUCHGC" check a bit more restrictively for valid values

### DIFF
--- a/mppsolar/protocols/pi30m045.py
+++ b/mppsolar/protocols/pi30m045.py
@@ -301,7 +301,7 @@ SETTER_COMMANDS = {
             b"(NAK\x73\x73\r",
             b"(ACK\x39\x20\r",
         ],
-        "regex": "MUCHGC(\\d\\d\\d\\d)$",
+        "regex": "MUCHGC(\\d[01]\\d[02])$",
     },
 }
 


mppsolar/protocols/pi30m045.py
```diff
@@ -301,7 +301,7 @@
             b"(NAK\x73\x73\r",
             b"(ACK\x39\x20\r",
         ],
-        "regex": "MUCHGC(\\d\\d\\d\\d)$",
+        "regex": "MUCHGC(\\d[01]\\d[02])$",
```
jblance, line 304: isnt this too restrictive? what the valid options for your inverter?

cordelster: Valid options are in increments of 10 from 10 to 120 except setting for 002 which is valid. 
I assume a majority will will adhere to increments of ten, though really this is probably not necessary as it will send back a NAK for most invalid values except where controller 0 does nothing and return ACK when used.
If this is possibly to restrictive for the possibility of other controllers, I understand. To me it's reference to whats considered valid values by the controller.

cordelster: My thought was I would rather see the error reflect in debug, rather that go out on the serial port to be returned by the controller.

PR #450, opened 2023-12-30
fixed crc calculation for axpert inverters

roysaad:
I noticed that sometimes the commands sent to my inverter (Axpert VM IV) timeout. I execute the QED[YYYYMMDD] command daily, many times per day to monitor the PV production per day and on some days every call will timeout. I tracked it to the CRC as one of the bytes or the CRC for those days equals 0x00. Incrementing this value to 0x01 will solve the problem.


jblance:
thanks for the contribution

PR #456, opened 2024-01-24
Powermon ved

jblance:


rossandrews (commented):
Sorry I disappeared for a bit, life got busy.

Glad to see you adding more protocols. Maybe add some tests for the protocol.

jblance:
> Sorry I disappeared for a bit, life got busy.
> 

no worries, life does that

> Glad to see you adding more protocols. Maybe add some tests for the protocol.

slowly adding protocols that work in a different way (also added templates, coz they still seem easiest way to re-format the raw data)

PR #497, opened 2024-06-23
fix pi17 GPMP0 regex to allow setting to all integers

rene-dev:
previously the default value of 10000 did not match the regex


jblance:
thanks for your contribution

Issue #112, opened 2021-05-08
JK BMS RS485 protocol


CRaccani:
Please find attached a different RS485 protocol document, sent to me by IC GOGOGO Store for the BMS I bought from them.
I have also just received their RS485 Dongle.
The Dongle plugs into the BMS but the dongle ends in 3 bare wires at the other end. I presume these are RS485- GND, A+, B+.
I have connected these to an RS485-USB dongle.
Any experience with this setup.

![20210508_212723](https://user-images.githubusercontent.com/64036712/117552627-81fc6a80-b044-11eb-9c81-965f43f403c6.jpg)

I tried your RS485 protocol but did not receive any bytes back. 
So looks like this BMS may use a different protocol.

Any chance of adding this protocol:
[RS485-english.pdf](https://github.com/jblance/mpp-solar/files/6446495/RS485-english.pdf)

jblance:
I've had a go at adding the first attempt at this protocol (called it jk232), likely still got work to do
can you update to version 0.7.54 or later and try with
`jkbms -p /dev/ttyUSB0 -b 9600 -P JK232 -I`
(change the port to whereever the adapter shows up), can you post the results

CRaccani:
Thank you. 
Here is the output requested:

$ jkbms -p /dev/ttyUSB0 -b 9600 -P JK232 -D
INFO:root:Solar Device Command Utility, version: 0.7.54, recent changes: add jk rs232/485 protocol
INFO:root:Creating device "unnamed" (type: "jkbms") on port "/dev/ttyUSB0 (porttype=None)" using protocol "JK232"
DEBUG:root:device_class <class 'mppsolar.devices.jkbms.jkbms'>
DEBUG:device:jkbms __init__ args ()
DEBUG:device:jkbms __init__ kwargs {'name': 'unnamed', 'port': '/dev/ttyUSB0', 'protocol': 'JK232', 'baud': 9600, 'porttype': None, 'mqtt_broker': 'localhost', 'mqtt_port': 1883, 'mqtt_user': None, 'mqtt_pass': None}
DEBUG:device:get_port_type: port matches ttyusb
INFO:device:set_port: Using serialio for communications
DEBUG:device:set_protocol: Protocol JK232
DEBUG:device:jkbms __init__ name unnamed, port <mppsolar.io.serialio.SerialIO object at 0xb6d609cc>, protocol <mppsolar.protocols.jk232.jk232 object at 0xb6d60a8c>
[...]

PR #508, opened 2024-08-18
Fixing my JK BMS RS485 issue.

lukolszewski:
This is to fix the issue described by my comment here https://github.com/jblance/mpp-solar/issues/506#issuecomment-2295194928 also a continuation of this issue https://github.com/jblance/mpp-solar/issues/112

It appears to be working fine for me.

The change adds a new serialio called jkserialio. The only difference from serialio is that instead of waiting for a CR it now waits for a specific byte sequence 68 00 00 XX XX (last 2 bytes are the CRC so can be anything). A better approach would be to process frame header and extract length info, but this was quicker to implement.

### DIFF
--- a/mppsolar/inout/__init__.py
+++ b/mppsolar/inout/__init__.py
@@ -11,6 +11,7 @@ class PortType(Enum):
     USB = auto()
     ESP32 = auto()
     SERIAL = auto()
+    JKSERIAL = auto()
     JKBLE = auto()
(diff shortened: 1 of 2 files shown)


jblance:
thanks for the contribution - not sure if this is needed (as dalyserial would solve the issue) or generic enough but i dont see it hurting (and mppsolar is a bit of a mess of bits anyway)

